PR #12, opened 2024-02-13
Implement title bar buttons for generic platforms like linux

cayasso:
This PR is preparation for removing frame and adding proper title bar for generic platforms such as Linux


davidmarkclements:
included in #110

PR #23, opened 2024-02-15
Add url parser

freeall:


lib/url.js
```diff
@@ -0,0 +1,42 @@
+'use strict'
+
+module.exports = parse
+
+function parse (url) {
+  const [ protocol, rest] = url.split('://')
```
mafintosh, line 6: use new URL so it validates the url

lib/url.js
```diff
@@ -0,0 +1,42 @@
...
+
+  if (protocol === 'pear') {
+    // pear://64CharLongKey
+    const hasCorrectLength = rest.length === 64
```
mafintosh, line 10: parse with hypercore-id-encoding

davidmarkclements:
completed via pear-link dep

PR #27, opened 2024-02-16
added --node flag to init cmd

rafapaezbas:


mafintosh:
killer

mafintosh (approved):

mattdesl:
there is a discrepancy, two modules exist but carry different versions on npm. is that intentional?

```
     'bare-fs': '^2.1.5',
     fs: 'npm:bare-node-fs',
```

mattdesl:
oh, confusing - all the dependencies are like that. there are two versions on npm of each? is that documented anywhere?

mafintosh:
documented here https://github.com/holepunchto/bare-node

mattdesl:
Perhaps an alias or import resolution map feature would be useful. The current approach seems prone to errors and lagging behind the base bare-x modules.

mafintosh:
yea on our radar, would set the process object also.

PR #38, opened 2024-02-19
use which-runtime module constants

rafapaezbas:


lib/constants.js
```diff
@@ -74,6 +63,8 @@ function toPath (url) {
 
 function electronModuleURL () {
   const u = require('url').pathToFileURL(process.execPath)
+  console.log(u)
```
davidmarkclements, line 66: stray logs

rafapaezbas: done

Issue #36, opened 2024-02-18
pear release production -> null [production]


serapath:
Somehow typing `pear release production` always spits out `null [production]`
Is that how it is supposed to be?

How can I see the latest version or the names i staged and released an app to?
`pear info` seems to be malformatted or maybe has missing information or maybe i just don't understand how to read it.

I can type `pear stage dev` in two different apps, but what if i move a folder or rename it. It doesn't seem to save any data in package.json or a `.pear` folder inside my app project, so how does it know about my project?

I think I broke something by renaming my project folder later on.

Clearer docs or some bug fixes would be helpful :-)

davidmarkclements:
@serapath the key for the app is based the app name+channel - the name of the app comes from package.json (but if you rename a folder and auto-update package.json name or something then it'll change the key) - but path shouldn't be part of it. For unusual circumstances you can override package.json name with the `--name` flag. `pear info` should be a lot clearer now with more info - `pear info <link>` will give you the channel name for an app. 

As for listing all channels that's an upcoming feature. Closing this as bugs are fixed, perpetually in agreement on clearer docs

PR #40, opened 2024-02-19
Fix unhandled error when missing release args and fix use of uninitialized name value

jkcdarunday:
## Changes:
- In both stage and release, `ctx.name` is used while it's still uninitialized causing `null` to be displayed as the app name (as described in #36). This fixes this by not using the context name and deriving the name directly.
- When running `pear release` without a specified key, a `Error: Cannot read properties of undefined (reading 'toString')` error is thrown instead of cleanly showing the "A key or the channel name must be specified." error. This fixes that by skipping the parsing when the passed argument is null/undefined/empty.

### DIFF
--- a/cmd/release.js
+++ b/cmd/release.js
@@ -16,7 +16,7 @@ module.exports = (ipc) => async function release (args) {
   try {
     const [from] = _
     let [, dir = ''] = _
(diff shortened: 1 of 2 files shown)


davidmarkclements:
superseded by next branch

PR #54, opened 2024-02-23
watch-drive for live-reload

davidmarkclements:


ipc/sidecar.js
```diff
@@ -127,10 +130,10 @@ module.exports = class IPC {
       if (messaged.has(app)) continue
       messaged.add(app)
 
-      if (appLink) {
-        if (appLink === app.bundle.link) {
-          app.notify({ type: 'pear/updates', app: true, version, diff: null })
-          app.message({ type: 'pear/updates', app: true, version, diff: null })
+      if (info.link && info.link === app.bundle.link) {
+        if (version.key === null || app.ctx.key?.hex === version.key) {
```
mafintosh, line 134: why is this if needed?

davidmarkclements: it's not, removing

PR #55, opened 2024-02-23
Add tests for Pear.updates

jkcdarunday:
## Blockers
- [ ] `pear release` does not trigger `Pear.updates`
    - Caused by `drive.watch()` only triggering on file changes and not drive.db writes ([see related lines](https://github.com/holepunchto/pear/blob/6b1c0b85961da114123dafd011d067f4f004b4eb/lib/bundle.js#L77-L83))
- [ ] `pear stage` on a released app triggers `Pear.updates` when it shouldn't
    - Caused by fetching release in #updates always returning null ([see related lines](https://github.com/holepunchto/pear/blob/6b1c0b85961da114123dafd011d067f4f004b4eb/lib/bundle.js#L78))


jkcdarunday:
This is already integrated into the tests

PR #56, opened 2024-02-23
Add tests for triggering Pear.updates when staging and releasing

jkcdarunday:
## Description
This contains only the working tests from https://github.com/holepunchto/pear/pull/55 rebased to the latest main

#### Current
- Add test for checking Pear.updates when staging and releasing
- Add test for checking Pear.updates when staging twice

#### Original changes that have been added to main:
- Create a separate test file for updates and call both smoke.test.js and updates.test.js from all.js
- Switch to stdout.write to work around console.logs in the test app not sending data until the app exits
- Fix error in pickMany where it deletes a non-existent key and causes a crash when an event is triggered twice

## Notes
- Not yet part of this are the tests for staging again after the first release as there's currently a bug where it calls Pear.updates when it shouldn't


test/03-updates.test.js
```diff
@@ -0,0 +1,178 @@
...
+  ok(seed3Announced, 'reseed announced')
+
+  comment('waiting for update')
+  const update2 = await inspector.awaitPromise(update2Promise.objectId)
+  const update2Version = update2?.value?.version
+  is(z32.encode(Buffer.from(update2Version?.key, 'hex')), key, 'app updated with matching key')
+  is(update2Version?.fork, 0, 'app version.fork is 0')
+  ok(update2Version?.length > update1Version?.length, `app version.length incremented (v${update2Version?.fork}.${update2Version?.length})`)
+
+  await inspector.evaluate('global.__PEAR_TEST__.inspector.disable()')
```
davidmarkclements, line 171: just do this automatically in inspector.close

test/03-updates.test.js
```diff
@@ -72,28 +63,22 @@ test('Pear.updates() should be called when restaging and releasing', async funct
...
-  const update2Version = update2?.value?.[1]?.version
-  is(z32.encode(Buffer.from(update2Version?.key, 'hex')), key, 'app update2 with matching key')
-  ok(update2Version?.length > update1Version?.length, 'app version incremented')
+  const update2 = await inspector.awaitPromise(update2Promise.objectId)
+  const update2Version = update2?.value?.version
+  is(z32.encode(Buffer.from(update2Version?.key, 'hex')), key, 'app updated with matching key')
+  is(update2Version?.fork, 0, 'app version.fork is 0')
+  ok(update2Version?.length > update1Version?.length, `app version.length incremented (v${update2Version?.fork}.${update2Version?.length})`)
 
   await inspector.evaluate('global.__PEAR_TEST__.inspector.disable()')
```
davidmarkclements, line 83: put this in inspector.close() (wrapped with a try/catch ignore)

test/03-updates.test.js
jkcdarunday: This line should now be in inspector._close

test/03-updates.test.js
jkcdarunday: This line should now be in inspector._close

PR #57, opened 2024-02-26
fix release watcher

mafintosh:


lib/bundle.js
```diff
@@ -291,5 +300,6 @@ module.exports = class Bundle {
     if (this.tracer) this.tracer.destroy()
     await this.drain()
     await this.drive.close()
+    if (this.watchingUpdates) this.watchingUpdates.destroy()
```
HDegroote, line 303: I think there's a race condition on the ready/close logic: if `close` is called before ready was finalised, the `#updates` logic might start setting up the watch logic after the check on `this.watchingUpdates` ran.
I'm not sure if that's bad, probably the `#updates` loop just crashes and exits, but maybe it can also hang forever, somehow waiting on an update event on a closed hyperdrive.

I think the clean solution to manage the lifecycle state is to use ready-resource for the `Bundle` class, replacing `#init` with `_open` and `close` with `_close`

mafintosh: Really needs to be an RR but out of scope for this pr, let me add a guard tho

HDegroote: yeah, agreed, I can PR the ready-resource after this one is

HDegroote (approved):
looks good to me (assuming the failing tests are expected)

mafintosh:
dont merge yet, hooking it up

PR #59, opened 2024-02-26
test teardown with pear-inspect

ruy-dan:
test teardown with pear-inspect


test/fixtures/terminal/index.js
```diff
@@ -1,32 +1,16 @@
...
+const inspector = new Inspector({ inspector: nodeInspector})
+const inspectorKey = await inspector.enable()
 
-if (debarg) {
-  const opts = debarg.split('=')[1].split(',')
-  debug.ready = opts.includes('ready')
-  debug.update = opts.includes('update')
-}
+stdout.write(`[ inspect ] key: ${inspectorKey.toString('hex')}\n`)
+stdout.write('[ inspect ] ready\n')
```
davidmarkclements, line 14: line isn't needed - ready is implied by having the key 

```
console.log({ key: inspectorKey.toString('hex') })
```

would be fine, no need for bare-pipe then

test/helper.js
```diff
@@ -102,6 +103,17 @@ class Helper {
       stdio: (global.Bare || global.process).argv.includes('--attach-boot-io') ? 'inherit' : 'ignore',
       cwd: this.platformDir
     })
+
+    sc.stdout.on('data', (data) => {
```
davidmarkclements, line 107: just inherit the stdio (data can be chunked right? so this basically amounts to "run stdout:" being interspersed amongst the chunks)

test/helper.js
```diff
@@ -110,29 +122,34 @@ class Helper {
...
     args = [...args, '--ua', 'pear/terminal']
 
-    const di = args.findIndex(arg => arg.startsWith('--debug='))
-    if (di > -1) args.push(args.splice(di, 1)[0])
-
     const iterable = new Readable({ objectMode: true })
 
-    const child = spawn(this.runtime, args, {
-      stdio: silent ? 'ignore' : ['inherit', 'pipe', 'pipe']
+    this.app = spawn(this.runtime, args, {
```
davidmarkclements, line 127: can you guarantee that the helper will only ever spawn one app at a time and `this.app` won't be overwritten by another spawn (thus losing the ref)?

davidmarkclements: do `yield { tag: 'child', child }`

test/helper.js
```diff
@@ -142,6 +159,10 @@ class Helper {
     yield * iterable
   }
 
+  closeApp (signal) {
+    this.app.kill(signal)
```
davidmarkclements, line 163: remove this

test/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "test",
   "version": "1.0.0",
   "description": "",
-  "main": "./smoke.test.js",
+  "main": "./teardown.test.js",
```
davidmarkclements, line 5: should this be all?

test/helper.js
```diff
@@ -110,29 +122,34 @@ class Helper {
...
-      if (str.indexOf('UPDATE') > -1) iterable.push({ tag: 'update', data })
+      if (str.indexOf('ready') > -1) {
+        iterable.push({ tag: 'ready', data })
+      }
+      if (str.indexOf('update') > -1) iterable.push({ tag: 'update', data })
+      if (str.indexOf('key') > -1) {
+        const match = str.match(/\[ inspect \] key: ([a-f0-9]+)\n/)
+        if (match) iterable.push({ tag: 'inspectkey', data: match[1] })
+      }
+      if (str.indexOf('teardown') > -1) iterable.push({ tag: 'teardown', data })
```
davidmarkclements, line 146: just do all this with the child directly outside of run

test/fixtures/terminal/index.js
ruy-dan: ready line removed

test/helper.js
ruy-dan: stdio set to 'inherit'

test/helper.js
ruy-dan: removed

test/package.json
ruy-dan: changed to all.js

scripts/pretest.js
```diff
@@ -43,8 +43,8 @@ const run = (cmd, args, opts) => {
   }
 
   if (isWindows) {
-    await run('pear.cmd', ['run', 'test'], { stdio: 'inherit', shell: true })
+    await run('pear.cmd', ['run', 'test', '--atach-boot-io'], { stdio: 'inherit', shell: true })
```
davidmarkclements, line 46: typo atach -> attach

test/smoke.test.js
```diff
@@ -3,58 +3,100 @@ const test = require('brittle')
...
 const os = require('bare-os')
 const Helper = require('./helper')
+const { Session } = require('pear-inspect')
+const { Readable } = require('streamx')
 
 test('smoke', async function ({ teardown, ok, is, plan, timeout, comment }) {
-  plan(3)
-  timeout(180000)
+  plan(4)
+  timeout(30000)
```
davidmarkclements, line 11: that's the default I think? if so remove timeout entirely

scripts/pretest.js
ruy-dan: fixed typo

test/fixtures/terminal/index.js
```diff
@@ -1,32 +1,19 @@
...
-     ▄▄▄▄▄▄▆▆▆▆▆▆       ${rst}${grn}pear
-    ▃▄▄▄▄▄▄▆▆▆▆▆▆▄      ${dim}${v(platform)}${rst}${grn}
-   ▄▄▄▄▄▄▄▄▆▆▆▆▆▆▆▆
-   ▄▄▄▄▄▄▄▄▆▆▆▆▆▆▆▆     ${rst}${grn}Welcome to the IoP
-     ▄▄▄▄▄▄▆▆▆▆▆▆
-       ▄▄▄▄▆▆▆▆
-`
-console.log('\n\x1b[s\x1b[J' + out + '\x1b[0m')
-if (debug.ready) console.log('[DEBUG] READY')
+stdout.write(inspectorKey.toString('hex'))
```
davidmarkclements, line 17: why not console.log ?

test/fixtures/terminal/index.js
```diff
@@ -1,32 +1,19 @@
...
-const argv = global.Bare?.argv || global.process.argv
-const debarg = argv.find(arg => arg.startsWith('--debug='))
+const inspector = new Inspector({ inspector: nodeInspector})
+const inspectorKey = await inspector.enable()
 
-if (debarg) {
-  const opts = debarg.split('=')[1].split(',')
-  debug.ready = opts.includes('ready')
-  debug.update = opts.includes('update')
+global.endInspection = async function () {
```
davidmarkclements, line 13: do
```
global.__PEAR_TEST__ = { inspector, inspectorKey } 
```

test/fixtures/terminal/index.js
```diff
@@ -1,32 +1,19 @@
 /* global Pear */
-const { config,versions } = Pear
-const [grn, rst, dim] = ['\x1b[32m', '\x1b[0m', '\x1b[2m']
-const v = ({ key, length, fork }) => `v${fork}.${length}.${(key += '').length <= 12 ? key : key.slice(0, 12) + '…'}`
-const { app, platform } = await versions()
+const { teardown } = Pear
+import nodeInspector from 'bare-inspector'
```
davidmarkclements, line 3: it's not `nodeInspector` though

davidmarkclements: also your package.json has it mapped to `inspector` so either remove that from export map and import bare-inspector or change it to import inspector

test/fixtures/terminal/index.js
ruy-dan: when running the app directly from the terminal, console.log logs as expected.However, when the app is run as a child process (not directly on a terminal) the logs are being buffered and not flushed to the parent until the child process exits. Bare Pipe, on the other hand, outputs directly.

test/fixtures/terminal/index.js
ruy-dan: removed imports map in package.json since we are only running tests in bare. renamed import to `bareInspector`

PR #67, opened 2024-03-04
(WIP) Treat updateNotify async

HDegroote:
The updateNotify logic seems to assume it's a sync function, but it's async, so the logic is bugged.

This is a WIP still because I wanted to sanity check first if the overall approach is still sound. The 2 TODO's indicate where it would still require changing


HDegroote:
The proper fix is to make updateNotify sync, here: https://github.com/holepunchto/pear/pull/68

PR #81, opened 2024-03-08
Trust dialog

rafapaezbas:


lib/engine.js
```diff
@@ -170,6 +170,7 @@ class Engine extends ReadyResource {
         if (report.type === 'update') return reports.update(report)
         if (report.type === 'upgrade') return reports.upgrade()
         if (report.type === 'restarting') return reports.restarting()
+        if (report.type === 'trust') return reports.trust(report)
```
davidmarkclements, line 173: name the report type and reports method `permissionRequired`

decal.html
```diff
@@ -694,6 +726,22 @@ <h1>Switch Keet back to ALPHA</h1>
...
               continue
             }
+
+            if (type === 'permissionRequired') {
+              const decalDialog = document.createElement('decal-dialog')
+              document.body.appendChild(document.createElement('pear-ctrl'))
+              document.body.appendChild(decalDialog)
+              decalDialog.setKey(report.key)
+              decalDialog.trust = async () => {
+                // await this.ipc.trust(report.key)
```
davidmarkclements, line 736:
```suggestion
                await this.ipc.trust({ key })
```

PR #83, opened 2024-03-12
Signal teardown for desktop apps

ruy-dan:
Listens for an exit signal (sigterm, sigint) and awaits the exit of clients on teardown. For desktop apps only, since terminal apps have this logic in the API already.


ruy-dan:
closing as discussed, as this is changing with the upcoming `pear-rpc` branch

PR #85, opened 2024-03-13
Fix URL handling on Windows and Linux

jkcdarunday:
## Changes
- Fix issues with the Linux desktop file
- Ensure that the executable path is up to date on Linux
- Replace electron-based URL registration (broken) on Windows with inserts to the registry
- Switch to the WAKEUP command instead of RUNTIME
- Added line to hide desktop entry from the user's application menu on Linux

## Notes
~~Since the pear-runtime.exe in Windows is a console application, running it spawns a console window before it launches the application window.~~ (fixed by switch to WAKEUP command)

### DIFF
--- a/electron-main.js
+++ b/electron-main.js
@@ -2,7 +2,7 @@
 const IPC = require('./ipc/main')
 const Context = require('./ctx/shared')
 const { App } = require('./lib/gui')
-const { SWAP, RUNTIME } = require('./lib/constants')
+const { SWAP, WAKEUP } = require('./lib/constants')
 const crasher = require('./lib/crasher')
 const connect = require('./lib/connect.js')
 const { isWindows, isMac, isLinux } = require('which-runtime')
@@ -42,11 +42,11 @@ async function electronMain () {
 function configureElectron () {
   const electron = require('electron')
   if (isLinux) {
-    linuxSetup(RUNTIME)
+    linuxSetup(WAKEUP)
   }
 
   if (isWindows) {
-    windowsSetup(RUNTIME)
+    windowsSetup(WAKEUP)
   }
 
   const appName = applingName()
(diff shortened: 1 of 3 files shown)


electron-main.js
```diff
@@ -114,9 +109,9 @@ function linuxSetup (executable) {
   const os = require('os')
   const { join } = require('path')
   const { spawnSync } = require('child_process')
-  const APP_NAME = 'Keet'
-  const ICON_NAME = 'keet'
-  const DESKTOP_FILE_NAME = 'keet.desktop'
+  const APP_NAME = 'Pear'
+  const ICON_NAME = 'pear'
+  const DESKTOP_FILE_NAME = 'pear.desktop'
```
jkcdarunday, line 114: @davidmarkclements Not too sure if this should be changed but having keet here didn't seem right.

electron-main.js
```diff
@@ -46,12 +46,7 @@ function configureElectron () {
   }
 
   if (isWindows) {
-    const ap = applingPath()
-    if (ap) {
-      electron.app.setAsDefaultProtocolClient('holepunch', ap) // legacy
-      electron.app.setAsDefaultProtocolClient('punch', ap) // legacy
-      electron.app.setAsDefaultProtocolClient('pear', ap)
-    }
+    windowsSetup(RUNTIME)
```
mafintosh, line 49: should be set to WAKEUP not runtime

electron-main.js
mafintosh: do we need to gc the keet.desktop file thats added here potentially?

jkcdarunday: Hmm.. My concern would be if something like the keet appimage creates a keet.desktop file, then we don't want to delete that.

Functionally, it should be alright if it exists since we set the handler to pear.desktop anyway but it does leave a Keet entry in the user's application list.

davidmarkclements: would need to do that in `migrate`, @jkcdarunday change sgtm

PR #90, opened 2024-03-15
sidecar startup log

ruy-dan:
Adds a log when sidecar has booted successfully to help with tests debugging.

### DIFF
--- a/test/helper.js
+++ b/test/helper.js
@@ -23,7 +23,6 @@ class Helper {
     this.teardown = t.teardown
     this.opts = opts
     this.logging = this.opts.logging
-    this.verbose = true
 
     this.client = null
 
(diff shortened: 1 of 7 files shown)


test/helper.js
```diff
@@ -89,6 +87,7 @@ class Helper {
     }
 
     function onconnect () {
+      if (this.logging) console.log('bootpipe connected to sidecar')
```
mafintosh, line 90: what does this.logging point to?

ruy-dan: true... this was pointing nowhere... the intention was to point to a property in the class helper, but `this` is not bound here. I simplified it to log always in tests for now

ruy-dan:
closing as this was merged into pear-rpc branch https://github.com/holepunchto/pear/pull/96

PR #95, opened 2024-03-20
print module import error for terminal apps

rafapaezbas:


mafintosh:
don't merge this until we know if this is fixed in the latest runtime builds

PR #98, opened 2024-03-22
Implement building app distributables using `pear build`

jkcdarunday:
## Introduction
Running `pear build <key> <build directory>` will:
1. Dump the app to the build directory
2. Generate a CMakeLists.txt from package.json if not included in the app code
3. Run `bare-dev configure` on the build directory
4. Run `bare-dev build` on the build directory

## Testing
- [x] Test on Linux
    - Requires `assets/linux/icon.png` to exist in the app
- [x] Test on Windows
    - Requires `assets/win32/icon.png` to exist in the app
    - Requires `assets/splash.png` to exist in the app
    - Requires Visual Studio 17 2022 to be installed
    - Requires to be run using x64 Native Tools Command Prompt for VS 2022
    - Requires [signing key to be generated](https://learn.microsoft.com/en-us/windows/msix/package/create-certificate-package-signing) and set in package.json
- [x] Test on Mac OS
    - Requires `assets/darwin/icon.png` to exist in the app
    - Requires `assets/splash.png` to exist in the app
    - Requires XCode utilities to be installed
    - Requires cmake to be installed (used brew)
    - Requires [signing key to be generated](https://support.apple.com/en-hk/guide/keychain-access/kyca8916/mac), [trusted](https://support.apple.com/en-hk/guide/keychain-access/kyca11871/mac), and set in package.json

## Other Notes
- (RESOLVED) Workarounds were needed to be able to run bare-dev and its `which` dependency:
   - Added aliases for `fs` and `path` in the dependencies
   - Set `global.process` to `require('bare-process')`
   - Set the `getuid` and `getgid` functions to return a hardcoded id in `global.process` as they are not currently implemented
- `pear build <channel>` is not yet supported as I couldn't find a way to resolve channel to key within the cli
- `pear dump <key>` seems to print file changes regardless of outputter

## Dependencies
- [x] Requires https://github.com/holepunchto/bare-dev/pull/11 to be merged and released


package.json
```diff
@@ -74,6 +77,7 @@
     "debounceify": "^1.1.0",
     "drive-bundler": "^1.0.0",
     "framed-stream": "^1.0.1",
+    "fs": "npm:bare-node-fs",
```
davidmarkclements, line 80: don't do this on platform - use the right thing in the right place and isBare from which-runtime module to determine in cases where it's both

davidmarkclements: (also, generally,  the preferred way to do this is import maps)

package.json
```diff
@@ -87,6 +91,7 @@
     "mime-db": "^1.52.0",
     "mirror-drive": "^1.3.0",
     "mutexify": "^1.4.0",
+    "path": "npm:bare-node-path",
```
davidmarkclements, line 94: remove

package.json
jkcdarunday: @davidmarkclements Ah, right now this is a workaround for the `which` dependency within bare-dev repo which is yet to be replaced.

https://holepunchdev.slack.com/archives/C054FE134P6/p1710861131438619?thread_ts=1710850156.778759&cid=C054FE134P6

davidmarkclements (changes requested):
just throw (then catch and print an error in cmd) if the dir isn't supplied  - see how it's done in dump

subsystems/build.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  init: require('bare-dev/init'),
```
davidmarkclements, line 2: are these explicitly marked as public apis by bare-dev or is this a workaround of missing API?

package.json
jkcdarunday: Tried using add the import map as well but it didn't work for bare-dev.

subsystems/build.js
jkcdarunday: Yup. These are directly exported by bare-dev in its package.json

cmd/build.js
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    print(`Building application ${key} for ${os.platform()}-${os.arch()}...`)
+
+    // TODO: Cleanup this workaround once pear-dev no longer relies on process
+    global.process = require('bare-process')
+    global.process.getuid = () => 1000
+    global.process.getgid = () => 1000
+
+    const drive = await createPlatformDrive()
```
davidmarkclements, line 36: is there a reason this is being done in the client? 
 
if not, then it needs to be implemented as  a method called `build` on the `Engine` class in lib/sidecar.js (pear-rpc branch). 

When the method is called `build` needs to load the system from the platform drive, which is on the engine instance (`this.drive`) 

You also need to add `build` to https://github.com/holepunchto/pear-ipc/blob/main/methods.js - it'll then be auto-added to the ipc instance so you can call it here as ipc.build(opts) 

See if you can also unload the subsystem when the client closes (in the build method on the Engine class in lib/sidecar.js)

jkcdarunday: @davidmarkclements Hmm.. I'm thinking that this could be problematic on Windows where the build process needs to execute in a specific environment (x64 Native Tools Command Prompt for VS 2022) otherwise it would not have the required commands such as cmake in its %PATH%.

If I understand it correctly, moving the build function to Engine would make it execute in sidecar which could potentially be running in a different environment (if it was previously started by running Keet etc.). It might lead to issues such as users trying to run pear build on the native tools command prompt where they have cmake but it keeps saying cmake not found.

package.json
jkcdarunday: Same as the other comment. `bare-dev` does not work and throws without this workaround (even with import map).

subsystems/build.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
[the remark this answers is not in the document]
jkcdarunday: I'm a bit confused about this one. Do you mean lib/build will only contain requires for bare-dev (as in this file) or are you referring to the code in `cli/build.js` instead?

jkcdarunday:
Rebased to the latest main branch

jkcdarunday:
Rebased to latest main and retested the bare-which changes on all platforms.

package.json
jkcdarunday: This should now be removed.

package.json
jkcdarunday: This should now be removed.

cmd/usage.js
```diff
@@ -151,6 +151,18 @@ module.exports = ({ fork, length, key }) => {
...
   `
 
+  const build = ansi.bold(cmd + ' build')
+  const buildArgs = ansi.bold('<key> <dir>')
+  const buildBrief = 'Build project distributable.'
+  const buildExplain = `${build} ${buildArgs}
+
+    ${buildBrief}
+
+    Specify the key to build a distributable for the project.
```
davidmarkclements, line 161:
```suggestion
    Creates a build folder at dir and creates a build for application key.
```

davidmarkclements (commented):
.

cmd/build.js
jkcdarunday: Update: this has now been removed as we've moved the call to the sidecar

cmd/usage.js
```diff
@@ -151,6 +151,18 @@ module.exports = ({ fork, length, key }) => {
...
   `
 
+  const build = ansi.bold(cmd + ' build')
+  const buildArgs = ansi.bold('<key> <dir>')
+  const buildBrief = 'Build project distributable.'
+  const buildExplain = `${build} ${buildArgs}
+
+    ${buildBrief}
+
+    Creates a build folder at dir and creates a build for application key.
```
davidmarkclements, line 161:
```suggestion
    Creates a a distributable from key per host machine at given dir.
```

cmd/usage.js
```diff
@@ -151,6 +151,18 @@ module.exports = ({ fork, length, key }) => {
...
   `
 
+  const build = ansi.bold(cmd + ' build')
+  const buildArgs = ansi.bold('<key> <dir>')
+  const buildBrief = 'Build project distributable.'
+  const buildExplain = `${build} ${buildArgs}
+
+    ${buildBrief}
+
+    Creates a a distributable from key per host machine at given dir.
```
davidmarkclements, line 161:
```suggestion
    Creates a a distributable per host OS at given dir from application key.
```

cmd/usage.js
```diff
@@ -151,6 +151,18 @@ module.exports = ({ fork, length, key }) => {
...
   `
 
+  const build = ansi.bold(cmd + ' build')
+  const buildArgs = ansi.bold('<key> <dir>')
+  const buildBrief = 'Build project distributable.'
+  const buildExplain = `${build} ${buildArgs}
+
+    ${buildBrief}
+
+    Creates a a distributable per host OS at given dir from application key.
```
davidmarkclements, line 161:
```suggestion
    Creates a distributable per host OS at given dir from application key.
```

PR #101, opened 2024-03-25
fix sidecar has clients

ruy-dan:


ruy-dan:
closing since a fix for this was commited to pear-rpc branch

PR #108, opened 2024-03-28
Pear info <channelName>

ruy-dan:
Updates pear info command with the following:

pear info [key]  -> pear info [key|channel]
pear info [key|channel] --changelog / pear info [key|channel] --no-changelog
pear info [key|channel] --full-changelog
pear info [key|channel] --metadata / pear info [key|channel] --no-metadata
pear info [key|channel] --key

### DIFF
--- a/cmd/usage.js
+++ b/cmd/usage.js
@@ -69,17 +69,23 @@ module.exports = ({ fork, length, key }) => {
   `
 
   const info = ansi.bold(cmd + ' info')
-  const infoArgs = ansi.bold('[key]')
+  const infoArgs = ansi.bold('<channel|key>')
   const infoBrief = 'Read project information.'
   const infoExplain = `${info} ${infoArgs}
 
     ${infoBrief}
 
-    Supply a key to view application info
+    Supply a key or channel to view application info
 
     Without a key pear info shows Pear info
 
-    --json          Newline delimited JSON output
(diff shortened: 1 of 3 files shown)


cmd/usage.js
```diff
@@ -69,17 +69,23 @@ module.exports = ({ fork, length, key }) => {
...
 
     ${infoBrief}
 
-    Supply a key to view application info
+    Supply a key or channel to view application info
 
     Without a key pear info shows Pear info
 
-    --json          Newline delimited JSON output
+    --json            Newline delimited JSON output
```
davidmarkclements, line 82: but the least important things near the end, put --json last

lib/sidecar.js
```diff
@@ -987,49 +987,73 @@ class Engine extends ReadyResource {
     await bundle.finalizeTracing()
   }
 
-  async * info ({ key } = {}, client) {
+  async * info ({ key, channel, dir, display } = {}, client) {
```
davidmarkclements, line 990: no to `display`, just match flags to keys as we do elsewhere - no nesting

cmd/usage.js
ruy-dan: moved json to the end

lib/sidecar.js
ruy-dan: removed `display` with nested props

PR #111, opened 2024-04-03
App storage move

davidmarkclements:


lib/sidecar.js
```diff
@@ -1133,24 +1133,30 @@ class Engine extends ReadyResource {
...
+      if (await exists(from) === false) {
         throw Object.assign(
-          new Error('App storage for ' + params.dst + ' already exists. Use --force to overwrite'),
-          { code: 'ERR_EXISTS' }
+          new Error('No app storage for found for ' + params.src),
+          { code: 'ERR_NOENT' }
         )
       }
+      if (await exists(to)) {
+        gc = path.join(GC, randomBytes(8).toString('hex'))
```
mafintosh, line 1145: you need to mkdir(GC) also if that folder doesn't exsist

PR #117, opened 2024-04-10
key format in Pear API update messages

ruy-dan:


ruy-dan:
According to docs,
1- `Pear.updates` should send a version object with a key that takes the format `<String(hex)>` [docs](https://github.com/holepunchto/pear-docs/blob/main/reference/pear/api.md#pearupdateslistener-async-functionfunction--streamxreadable)
2- `Pear.versions` should send a key with format `<Buffer>` (https://github.com/holepunchto/pear-docs/blob/main/reference/pear/api.md#pearversions-object)
3- `Pear.config` should send a key as an object with format `{ z32: <String>, hex: <String>},`(https://github.com/holepunchto/pear-docs/blob/main/reference/pear/api.md#pearconfig-object)


At the moment, 
1- `Pear.updates` sends a key with different format depending on if it's an App update or a Platform update: 
- Platform updates: `<String(z32)>`
- App updates: `<String(hex)>`

2- `Pear.versions` sends  an object that contains a key with format:
- Platform updates: `<String(z32)>` or `String(path_to_repo)` if we are in localdev
- App updates: `<String(z32)>`

davidmarkclements:
closing for backlog

PR #118, opened 2024-04-10
add general copy/paste options using right-click

yassernasc:
Following the changes:

- removed ContextMenu's `x` and `y` attrs, since isn't used, apparently
- refactored the logic to check if the content to be copied is an image
- besides the basic copy and paste, I've added an extra `Copy Link` option
- because the accused behavior "on Linux it literally say "(empty)", was added a check to verify if it would have options to be shown before actually creating the Menu Popup, at the moment I can't check if this detail is resolved

### DIFF
--- a/gui/gui.js
+++ b/gui/gui.js
@@ -302,18 +302,44 @@ class ContextMenu {
   constructor (webContents, { devtools } = {}) {
     this.webContents = webContents
     this.devtools = devtools
-    this.x = 0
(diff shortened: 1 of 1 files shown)


rafapaezbas:
Tested in Linux:

Screenshot from 2024-04-11 13-29-00

rafapaezbas:
Tested in Linux :+1:

PR #119, opened 2024-04-11
Run CI on any PR, and run CI on push to next

HDegroote:


HDegroote:
To be useful, this commit should then be merged in next as well (currently the CI does not run there)

HDegroote:
Mac CI failed, but that's unrelated to the changes to the CI file here, which should be safe to merge (it only changes when the CI runs, not how it runs)

PR #122, opened 2024-04-11
copy/paste text only using right-click

yassernasc:
Changes from the last PR:

- Check `user-select` css prop to avoid copying elements with `user-select: none;`
- Fix paste behavior on buttons with a more strick check, line  (can't reproduce this bug, my test cases are much more modest than the entire Keet application, but I am confident that it's fixed)
- Fix the "Inspect Element" option, the bug was caused by a singleton state don't being updated, my fix was passing the state as a parameter

### DIFF
--- a/gui/gui.js
+++ b/gui/gui.js
@@ -299,13 +299,12 @@ function refresh (webContents, ctx) {
 }
 
 class ContextMenu {
-  constructor (webContents, { devtools } = {}) {
+  constructor (webContents) {
     this.webContents = webContents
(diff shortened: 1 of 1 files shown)


yassernasc:
After feedback, I've tried to keep to the PR as simple as possible by:

- copy/paste text only
- `user-select` check is no more needed
- removed the "Copy Link", it was a bonus thing and also would need the `user-select` check

yassernasc:
re-pushed, 'cause I forgot to run lint

PR #125, opened 2024-04-15
Clean up errors

rafapaezbas:


gui/gui.js
```diff
@@ -837,7 +845,7 @@ class GuiCtrl {
   }
 
   async send (...args) {
-    if (this.closed) throw Error(`Cannot send to closed ${this[kCtrl]}`)
+    if (this.closed) throw CANNOT_SEND(`Cannot send to closed ${this[kCtrl]}`)
```
mafintosh, line 848: this one should be INSTANCE_CLOSED or similar

gui/gui.js
```diff
@@ -877,7 +885,7 @@ class GuiCtrl {
   }
 
   async getMediaSourceId () {
-    if (this.closed) throw Error(`Cannot get media source id if the ${this[kCtrl]} is closed`)
+    if (this.closed) throw CANNOT_GET_MEDIA_SOURCE_ID(`Cannot get media source id if the ${this[kCtrl]} is closed`)
```
mafintosh, line 888: same as above

gui/gui.js
rafapaezbas: done

gui/gui.js
rafapaezbas: done

cmd/iface.js
```diff
@@ -60,19 +60,20 @@ function indicator (value, type = 'success') {
...
       if (json) {
         print(JSON.stringify({ cmd, tag, data }))
         continue
       }
-      if (tag === 'error') {
-        const err = new Error(data.message)
-        err.sidecarCode = data.code
-        err.sidecarStack = data.stack
-        throw err
+      let result = null
```
davidmarkclements, line 70: have you checked that this change doesn't cause breakage for other command outputs? it definitely will if any of them push objects with tag: 'error' 

what's the reason for this change?

rafapaezbas: sorry, that was unintended, let me revert

PR #128, opened 2024-04-16
added silent flag for detached run

rafapaezbas:
This PR adds a --silent flag that fixes the detached process crashing due to the write the the std i/o of a parent process already ended.


rafapaezbas:
its enough if we check for the --detach flag

PR #129, opened 2024-04-16
ignore stdio for --detached run

rafapaezbas:


lib/run.js
```diff
@@ -22,12 +22,11 @@ module.exports = async function run ({ ipc, args, key, storage, detached, dir })
       }
 
       args = args.filter((arg) => arg !== '--detached')
-      const opts = { detached: true }
+      const opts = { detached: true, stdio: 'ignore' }
```
davidmarkclements, line 25: is this not implied by detached: true?

rafapaezbas: no, it defaults to `pipe`

PR #131, opened 2024-04-18
pear gc sidecar

ruy-dan:


lib/sidecar.js
```diff
@@ -1390,4 +1412,96 @@ class Engine extends ReadyResource {
...
+  async #kill (ids) {
+    if (ids.length === 0) return []
+
+    const command = isWindows ? 'taskkill' : 'kill'
+    const args = isWindows ? ['/F', '/PID'].concat(ids) : ids
+
+    return new Promise((resolve, reject) => {
+      const sp = spawn(command, args)
+
+      sp.stderr.on('data', data => { console.log(data.toString()) })
```
davidmarkclements, line 1472: errors a logged to stderr (console.error) not stdout (console.log) 

also, if you want to output the error output in sidecar process from a spawn just inherit STDERR - but only if verbose

lib/sidecar.js
```diff
@@ -1390,4 +1412,96 @@ class Engine extends ReadyResource {
...
+  async #lookupProcess ({ name = null, flag = null } = {}) {
+    if (!name || !flag) throw new Error('Process name or flag are missing')
+
+    const [sh, args] = isWindows
+      ? ['cmd.exe', ['/c', `wmic process where "executablepath like '%${name}%'" get processid, commandline`]]
+      : ['/bin/sh', ['-c', `ps ax | grep -i -- '${name}' | grep -i -- '${flag}'`]]
+
+    return new Promise((resolve, reject) => {
+      const sp = spawn(sh, args)
+      const output = []
```
davidmarkclements, line 1433: there's no need to buffer the output - send it over the wire immediataley by yielding it from the parent - easiest way is to make it so Collector sidecar returns a stream instead of a promise and then yield the tags { tag: 'stderr, data ... }

PR #132, opened 2024-04-18
hide minimize/maximize buttons from trust dialog

rafapaezbas:


rafapaezbas:
To be tested on windows and Mac

preload.js
```diff
@@ -159,6 +161,37 @@ if (process.isMainFrame) {
...
+      })
+    }
+
+    async setCtrl ({ maximize = true, minimize = true }) {
+      if (maximize === true) {
+        const max = this.root.querySelector('#max')
+        max.style.display = 'inline'
+        if (isMac) await gui.ipc.setMaximizable({ id: gui.id, value: true })
+      }
+      if (maximize === false) {
```
HDegroote, line 180: Don't think there's any need for the if-else construct here and the one below, so easier to maintain without them:

```
        const min = this.root.querySelector('#min')
        min.style.display = minimize ? 'inline' : 'none'
        if (isMac) await gui.ipc.setMinimizable({ id: gui.id, value: minimize })
```

gui/gui.js
```diff
@@ -1586,6 +1596,16 @@ class PearGUI extends ReadyResource {
 
   setSize ({ id, width, height }) { return this.get(id).setSize(width, height) }
 
+  setMinimizable ({ id, value }) {
+    this.message({ type: 'pear/setMinimize', data: value })
+    return this.get(id).setMinimizable(value)
```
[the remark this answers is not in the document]
davidmarkclements: @HDegroote no the `get` method returns a "ghost" object with these funcs - what we should do at another point though is log out if the ghost ever gets used (because it shouldn't, it's backstop)

preload.js
```diff
@@ -159,6 +176,29 @@ if (process.isMainFrame) {
...
+      this.#setCtrl()
+    }
+
+    async #setCtrl () {
+      if (this.dataset.minimizable !== undefined) this.#setMinimizable(strToBool(this.dataset.minimizable))
+      if (this.dataset.maximizable !== undefined) this.#setMaximizable(strToBool(this.dataset.maximizable))
+    }
+
+    async #setMaximizable (value) {
+      if (!isMac) {
```
davidmarkclements, line 189: if (isMac) - no need to not it

rafapaezbas:
> To be tested on windows and Mac

tested

PR #134, opened 2024-04-18
filter Linux Electron error logs

rafapaezbas:


lib/run.js
```diff
@@ -117,7 +117,9 @@ module.exports = async function run ({ ipc, args, key, storage, detached, dir })
           const ignore = str.indexOf('DevTools listening on ws://') > -1 ||
                 str.indexOf('NSApplicationDelegate.applicationSupportsSecureRestorableState') > -1 ||
                 str.indexOf('devtools://devtools/bundled/panels/elements/elements.js') > -1 ||
-                str.indexOf('sysctlbyname for kern.hv_vmm_present failed with status -1') > -1
+                str.indexOf('sysctlbyname for kern.hv_vmm_present failed with status -1') > -1 ||
```
HDegroote, line 120: Not related to this PR, since that's just following the style of the surrounding code, so no need to address these before merging. But I have some remarks on this code block in general:

- I don't think the data chunks emitted by stderr/stdout are guaranteed to never be split halfway through one of those strings, nor that it can contain more than a single line (for example an actual error and one which should be ignored living in the same data chunk). A newline-decoder can be used to address this (example here: https://github.com/holepunchto/hyperdht/blob/225eaaadb8f7fca57efb967e86d5b8538d0127b6/test/helpers/index.js#L26-L28). Since it's complicated to reason about regardless, because of all the different OS's this might run in, I think a newline decoder is a sane default

- I think `str.includes(...)` is cleaner than `str.indexOf(...) > -1

- Having a list of strings to ignore would be cleaner IMO, then you can just do something like ` const ignore = ignoreStrings.every(d => !str.includes(d))`

davidmarkclements: @HDegroote yeah as we talked about this can be an incremental improvement, in practice not an urgent concern

davidmarkclements:
failing linux ci unrelated

PR #143, opened 2024-04-30
Windows pipe name

ruy-dan:
Updates the pipe name on Windows by deriving it from the platform dir so that each platform dir is associated with a unique pipe. This aligns with our current mac/linux implementation, where different sockets are used for different platform instances.

### DIFF
--- a/lib/constants.js
+++ b/lib/constants.js
@@ -2,6 +2,8 @@
 const HypercoreID = require('hypercore-id-encoding')
 const { platform, arch, isWindows, isLinux } = require('which-runtime')
 const { pathToFileURL, fileURLToPath } = require('url-file-url')
+const sodium = require('sodium-native')
+const b4a = require('b4a')
 
 const BIN = 'by-arch/' + platform + '-' + arch + '/bin/'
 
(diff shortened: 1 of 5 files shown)


lib/constants.js
```diff
@@ -20,6 +22,10 @@ const PLATFORM_URL = LOCALDEV ? new URL('pear/', swapURL) : new URL('../../../',
 const PLATFORM_DIR = toPath(PLATFORM_URL)
 const PLATFORM_LOCK = toPath(new URL('corestores/platform/primary-key', PLATFORM_URL))
 
+const buf = b4a.allocUnsafe(32)
+sodium.crypto_generichash(buf, b4a.from(PLATFORM_DIR))
+const PIPE_ID = b4a.toString(buf, 'hex')
```
mafintosh, line 27: i would define a helper in the bottom of the file hashToHex to make this a little clearer. then we can also "save" the compute on none windows, ie

```
const PIPE_ID = isWindows ? hashToHex(...) : null
```

mafintosh (approved):

PR #154, opened 2024-05-15
Fix pear restart by passing and using clientArgv

jkcdarunday:


subsystems/sidecar/index.js
```diff
@@ -475,32 +480,64 @@ class Sidecar extends ReadyResource {
...
+        }
+
+        spawn(RUNTIME, cmdArgs, opts).unref()
       }
 
       return
     }
 
-    const restarts = await this.#shutdown(client)
+    const sidecarClosed = new Promise(resolve => this.corestore.on('close', resolve))
```
davidmarkclements, line 504:
```suggestion
    const sidecarClosed = new Promise((resolve) => this.corestore.on('close', resolve))
```

PR #157, opened 2024-05-17
Print error on release if 'previewFor' is present

ruy-dan:


lib/sidecar.js
```diff
@@ -1104,6 +1105,10 @@ class Release extends OpStream {
       throw ERR_RELEASE_ERROR(`The "${name}" app has not been staged on ${channel ? '"' + channel + '" channel' : link}.`)
     }
 
+    if (ctx.config.options.previewFor !== undefined) {
+      throw ERR_INVALID_CONFIG('Config contains a \'previewFor\' property. This feature is only valid for pre-release.')
```
davidmarkclements, line 1109:
```suggestion
      throw ERR_INVALID_CONFIG('The `pear.previewFor` package.json field is unuseful and invalid for production releases. Remove the field in order to release. ')
```

PR #163, opened 2024-05-22
bypass electron media access status in Linux and ask for media access in linux and windows

rafapaezbas:
`getMediaAccessStatus` isn't supported by Electron on Linux.

- Mac: Allows programmatic access requests.
- Windows: Does not support programmatic access requests; must be done via the Windows user settings UI. However, it does allow checking the status.
- Linux: Lacks any standard for requesting or checking media access status.

The pull request also corrects a bug introduced in the line: `desktopSources: (options = null) => ipc.desktopSources(options = null)`

Bug introduced in commit https://github.com/holepunchto/pear/commit/2fcfad16b6092fe099b9df96ba3cf8db012464b7

### DIFF
--- a/gui/gui.js
+++ b/gui/gui.js
@@ -1497,7 +1497,7 @@ class PearGUI extends ReadyResource {
 
   getMediaAccessStatus ({ mediaType }) {
     if (isLinux) {
-      return 'unknown'
+      return 'unsupported'
     } else {
       return electron.systemPreferences.getMediaAccessStatus(mediaType)
     }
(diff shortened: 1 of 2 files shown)


gui/gui.js
```diff
@@ -1496,11 +1496,15 @@ class PearGUI extends ReadyResource {
   }
 
   getMediaAccessStatus ({ mediaType }) {
-    return electron.systemPreferences.getMediaAccessStatus(mediaType)
+    if (isLinux) {
+      return 'unknown'
```
davidmarkclements, line 1500:
```suggestion
      return 'unsupported'
```

Issue #91, opened 2024-03-15
pear update/restart on a console app not working


ryanramage:
here is the simplest pear app to reproduce 

```
/* global Pear */
console.log(`welcome!`)

Pear.updates((info) => {
  console.log('got update', info)
  Pear.restart()
})
```

If you run this with its key, and do pear stage dev, this is what keeps happening. sometime it crashes or sometimes it just hangs, without reload. but always it dumps out the exception seen below

```
~/apps/secondpear-term via  v18.14.1 took 8s
❯ pear run pear://zrdk7f7mf9xwyj7e1gfeyg5haqryuqxwkykf9w6f9twtgxqayyqo
welcome!
got update {
  type: 'pear/updates',
  app: true,
  version: {
    key: 'b906ae97ab2fdf4027a8918a801b7cc38809b9f450145fd3c5fc69133dd8001d',
    length: 1400,
    fork: 0
  },
  diff: null
}
2024-03-15T17:52:59.003Z cli exiting due to unhandled rejection
key=pqbzjhqyonxprx8hghxexnmctw75mr91ewqw5dxe1zmntfyaddqy
length=2727
fork=0
platform=darwin
arch=arm64
pid=37248
uptime=16.024s
stack=Error: [E_MUX_REMOTE] JSONRPC-MUX: channel remotely closed
    at JSONRPCMuxChannel.close (file:///Users/ryan/Library/Application%20Support/pear/by-dkey/7216eeac5c879a2ca1dcff14bb8c1ad47d188a8c17b13e66f36d354ea8bd3b28/2/boot.bundle/node_modules/jsonrpc-mux/index.js:76:13)
    at Channel.onclose (file:///Users/ryan/Library/Application%20Support/pear/by-dkey/7216eeac5c879a2ca1dcff14bb8c1ad47d188a8c17b13e66f36d354ea8bd3b28/2/boot.bundle/node_modules/jsonrpc-mux/index.js:21:20)
    at Channel._close (file:///Users/ryan/Library/Application%20Support/pear/by-dkey/7216eeac5c879a2ca1dcff14bb8c1ad47d188a8c17b13e66f36d354ea8bd3b28/2/boot.bundle/node_modules/protomux/index.js:142:22)
[...]

PR #183, opened 2024-06-05
Add pear terminal restart

jkcdarunday:
Requires https://github.com/holepunchto/pear-runtime-bare/pull/2 to be merged and released first for this to work

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -95,7 +95,7 @@ class API {
 
   versions = () => this.#reftrack(this.#ipc.versions())
 
-  restart = (opts = {}) => {
+  restart = async (opts = {}) => {
     // TODO: use Pear.shutdown when it lands instead
     if (this.#state.type === 'terminal' && !opts.platform) Bare.exit(BARE_RESTART_EXIT_CODE)
     else return this.#reftrack(this.#ipc.restart(opts))
--- a/lib/api.js
+++ b/lib/api.js
@@ -1,5 +1,6 @@
 'use strict'
 const Worker = require('./worker')
+const { BARE_RESTART_EXIT_CODE } = require('../constants')
 const noop = () => {}
 const teardown = global.Bare ? require('./teardown') : noop
 const program = global.Bare || global.process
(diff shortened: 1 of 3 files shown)


run/index.js
```diff
@@ -37,7 +38,11 @@ module.exports = async function run ({ ipc, args, cmdArgs, link, storage, detach
     dir = path.resolve(cwd, dir)
   }
 
-  if (dir !== cwd) os.chdir(dir)
+  if (dir !== cwd) {
+    Bare.on('exit', () => os.chdir(cwd)) // TODO: Remove this once teardown is fixed
```
davidmarkclements, line 42:
```suggestion
    Bare.on('exit', () => os.chdir(cwd)) // TODO: remove this once Pear.shutdown is used to close
```

lib/api.js
```diff
@@ -94,7 +95,10 @@ class API {
 
   versions = () => this.#reftrack(this.#ipc.versions())
 
-  restart = (opts = {}) => this.#reftrack(this.#ipc.restart(opts))
+  restart = (opts = {}) => {
+    if (this.#state.type === 'terminal' && !opts.platform) Bare.exit(BARE_RESTART_EXIT_CODE)
```
davidmarkclements, line 99:
```suggestion
    // TODO: use Pear.shutdown when it lands instead & make restart async
    if (this.#state.type === 'terminal' && !opts.platform) Bare.exit(BARE_RESTART_EXIT_CODE)
```

lib/api.js
```diff
@@ -94,7 +95,11 @@ class API {
 
   versions = () => this.#reftrack(this.#ipc.versions())
 
-  restart = (opts = {}) => this.#reftrack(this.#ipc.restart(opts))
+  restart = (opts = {}) => {
+    // TODO: use Pear.shutdown when it lands instead & make restart async
```
davidmarkclements, line 99:
```suggestion
    // TODO: use Pear.shutdown when it lands instead
```

lib/api.js
```diff
@@ -94,7 +95,11 @@ class API {
...
+    if (this.#state.type === 'terminal' && !opts.platform) Bare.exit(BARE_RESTART_EXIT_CODE)
+    else this.#reftrack(this.#ipc.restart(opts))
```
davidmarkclements, line 101:
```suggestion
    else return this.#reftrack(this.#ipc.restart(opts))
```

davidmarkclements:
note: does not require pear-runtime-bare build to actually be merged, it just means we start with hard restart, exit 75, it's on user system to restart. This is valid. Once pear-runtime-build lands, it'll automatically restart in place, switching to a soft exit - with an option to specify hard exit for full process restart (this will lose I/O whereas soft restart won't). 

first PR towards closing https://github.com/holepunchto/pear/issues/91

up next will be full platform restart support, both hard/soft

PR #187, opened 2024-06-11
Add support for Pear platform restarts in terminal apps

jkcdarunday:
Adds support for:
- `Pear.restart({ platform: true })` which tells the terminal app to restart in-place (soft restart)
- `Pear.restart({ platform: true, hard: true })` which completely relaunches terminal apps through sidecar instead restarting in-place

### DIFF
--- a/run/index.js
+++ b/run/index.js
@@ -114,17 +114,17 @@ module.exports = async function run ({ ipc, args, cmdArgs, link, storage, detach
 
     const pear = new API(ipc, state)
 
-    pear.messages({ type: 'pear/restart' }, async () => {
+    global.Pear = pear
+
+    Pear.messages({ type: 'pear/restart' }, async () => {
       ipc.stream.destroy()
 
       const fd = await new Promise((resolve, reject) => fs.open(PLATFORM_LOCK, 'r+', (err, fd) => err ? reject(err) : resolve(fd)))
       await fsext.waitForLock(fd)
       await new Promise((resolve, reject) => fs.close(fd, (err) => err ? reject(err) : resolve(fd)))
 
-      await pear.restart()
+      await Pear.restart()
     })
-
-    global.Pear = pear
 
     const protocol = new Module.Protocol({
       exists (url) {
--- a/run/index.js
+++ b/run/index.js
@@ -34,6 +34,7 @@ module.exports = async function run ({ ipc, args, cmdArgs, link, storage, detach
   }
 
   let cwd = os.cwd()
+  const originalCwd = cwd
   let dir = cwd
   let base = null
   if (key === null) {
@@ -43,8 +44,8 @@ module.exports = async function run ({ ipc, args, cmdArgs, link, storage, detach
     base = project(dir, pathname, cwd)
     dir = base.dir
     if (dir !== cwd) {
-      Bare.on('exit', () => os.chdir(cwd)) // TODO: remove this once Pear.shutdown is used to close
-      teardown(() => os.chdir(cwd))
+      Bare.on('exit', () => os.chdir(originalCwd)) // TODO: remove this once Pear.shutdown is used to close
+      teardown(() => os.chdir(originalCwd))
       os.chdir(dir)
       cwd = dir
     }
--- a/run/index.js
+++ b/run/index.js
@@ -117,7 +117,8 @@ module.exports = async function run ({ ipc, args, cmdArgs, link, storage, detach
 
     global.Pear = pear
 
-    global.Pear.messages({ type: 'pear/reload' }, async () => {
+    const reloadSubscriber = ipc.messages({ type: 'pear/reload' })
+    reloadSubscriber.on('data', async () => {
       ipc.stream.destroy()
 
       const fd = await new Promise((resolve, reject) => fs.open(PLATFORM_LOCK, 'r+', (err, fd) => err ? reject(err) : resolve(fd)))
--- a/run/index.js
+++ b/run/index.js
@@ -20,7 +20,8 @@ const {
 const parseLink = require('./parse-link')
 const teardown = require('../lib/teardown')
 const { isWindows } = require('which-runtime')
-const { BARE_RESTART_EXIT_CODE } = require('../constants')
+const { PLATFORM_LOCK } = require('../constants')
+const fsext = require('fs-native-extensions')
 
 module.exports = async function run ({ ipc, args, cmdArgs, link, storage, detached, flags, appArgs, indices }) {
   const { drive, pathname } = parseLink(link)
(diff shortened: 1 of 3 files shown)


run/index.js
```diff
@@ -114,6 +115,15 @@ module.exports = async function run ({ ipc, args, cmdArgs, link, storage, detach
     state.update({ config: await ipc.config() })
 
     const pear = new API(ipc, state)
+
+    pear.messages({ type: 'pear/restart' }, async () => {
```
davidmarkclements, line 119: this is what pear.restart() is supposed to do

run/index.js
```diff
@@ -114,6 +115,15 @@ module.exports = async function run ({ ipc, args, cmdArgs, link, storage, detach
...
+      // Wait for sidecar to shutdown
+      // TODO: Figure out how to properly detect sidecar shutdown from the client
+      await new Promise(resolve => setTimeout(resolve, 1500))
```
davidmarkclements, line 122: https://github.com/holepunchto/pear-ipc/blob/main/api.js - add restart to api.js and use the lock file approach - then just called Pear.restart

run/index.js
```diff
@@ -114,6 +115,15 @@ module.exports = async function run ({ ipc, args, cmdArgs, link, storage, detach
...
+
+      global.Bare.exit(BARE_RESTART_EXIT_CODE)
```
davidmarkclements, line 124: Pear.restart should be doing this

subsystems/sidecar/index.js
```diff
@@ -470,24 +470,38 @@ class Sidecar extends ReadyResource {
 
   shutdown (params, client) { return this.#shutdown(client) }
 
-  closeClients () {
+  get uniqueClients () {
```
davidmarkclements, line 473: no getter, just make clients unique inline

subsystems/sidecar/index.js
```diff
@@ -538,15 +563,15 @@ class Sidecar extends ReadyResource {
...
       const opts = { cwd: dir, env, detached: true, stdio: 'ignore' }
       if (appling) {
         const applingPath = typeof appling === 'string' ? appling : appling?.path
         if (isMac) spawn('open', [applingPath.split('.app')[0] + '.app'], opts).unref()
         else spawn(applingPath, opts).unref()
       } else {
-        // TODO: TERMINAL_RUNTIME restarts
-        const RUNTIME = this.updater === null ? DESKTOP_RUNTIME : this.updater.swap + DESKTOP_RUNTIME.slice(SWAP.length)
+        const baseRuntime = options?.type === 'terminal' ? RUNTIME : DESKTOP_RUNTIME
+        const runtime = this.updater === null ? baseRuntime : this.updater.swap + baseRuntime.slice(SWAP.length)
```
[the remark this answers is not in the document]
davidmarkclements: also inline baseRuntime into the ternary - 2 level ternary is fine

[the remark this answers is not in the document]
jkcdarunday: Without renaming, RUNTIME (non-desktop runtime) from import becomes inaccessible here I believe due to hoisting

subsystems/sidecar/index.js
```diff
@@ -527,8 +541,19 @@ class Sidecar extends ReadyResource {
...
     }
 
+    if (!hard) {
+      const clients = this.uniqueClients.filter(c => c?.userData?.state?.options?.type === 'terminal')
+      if (this.verbose) console.log(`Soft-restarting ${clients.length} terminal app(s)`)
+      for (const client of clients) client.userData.message({ type: 'pear/restart' })
+
+      // Wait for 'pear/restart' messages to send and get handled by clients
+      // TODO: Figure out how to cleanly flush this buffer
+      await new Promise(resolve => setTimeout(resolve, 1500))
```
davidmarkclements, line 551: don't understand why this is needed, but if it is you'll need an ipc call from client a counter - but ideally don't do this at all, just wait for sidecar to shutdown (can't be any clients if it's shut) and then trigger restarts

run/index.js
davidmarkclements: sorry scrap that this part is fine

run/index.js
davidmarkclements: this is client side, so it should be calling Pear.restart

subsystems/sidecar/index.js
```diff
@@ -537,15 +557,17 @@ class Sidecar extends ReadyResource {
 
     await sidecarClosed
 
-    for (const { dir, appling, cmdArgs, env } of restarts) {
+    const TERMINAL_RUNTIME = RUNTIME
```
davidmarkclements, line 560: remove this, just use RUNTIME

subsystems/sidecar/index.js
```diff
@@ -526,8 +526,28 @@ class Sidecar extends ReadyResource {
...
+      if (this.verbose && terminalClients.length > 0) {
+        console.log(`Soft-restarting ${terminalClients.length} terminal app(s)`)
+      }
+      for (const client of terminalClients) client.userData.message({ type: 'pear/restart' })
+    }
+
     const sidecarClosed = new Promise((resolve) => this.corestore.once('close', resolve))
-    const restarts = (await this.#shutdown(client)).filter(({ run }) => run)
+    const restarts = (await this.#shutdown(client))
+      .filter(({ run, options }) => run && (options?.type !== 'terminal' || hard))
```
davidmarkclements, line 550: don't to || hard - just don't filter if hard

subsystems/sidecar/index.js
```diff
@@ -526,8 +526,28 @@ class Sidecar extends ReadyResource {
...
+
+        terminalClients.push(client)
+      }
+
+      if (this.verbose && terminalClients.length > 0) {
+        console.log(`Soft-restarting ${terminalClients.length} terminal app(s)`)
+      }
+      for (const client of terminalClients) client.userData.message({ type: 'pear/restart' })
+    }
+
```
davidmarkclements, line 547: all of this should be no more than 5 LOC and just the one loop. do `const restarts = ... filter(...)` as a few lines below.

davidmarkclements: also remove the log here - instead just put (hard) or (soft) in the console.log at the top of the restart function

davidmarkclements (changes requested):
feedback + we also need soft restart for platform: false - e.g. a client should be about to soft restart itself

subsystems/sidecar/index.js
davidmarkclements: let's rename pear/restart to pear/reload and send this to all clients, not just terminal clients

subsystems/sidecar/index.js
jkcdarunday: Tried this again. Same as before. It crashes when using RUNTIME when declaring RUNTIME in line 568 which I believe is due to hoisting.

davidmarkclements: ok rename RUNTIME on 558 to TARGET_RUNTIME instead

subsystems/sidecar/index.js
```diff
@@ -526,8 +526,18 @@ class Sidecar extends ReadyResource {
       return
     }
 
+    if (!hard && this.hasClients) {
+      const seen = new Set()
+      for (const { userData: app } of this.client) {
+        if (!app.state || seen.has(app.state.id)) return
```
davidmarkclements, line 532: should be `continue`?

davidmarkclements (approved):

run/index.js
```diff
@@ -107,8 +109,20 @@ module.exports = async function run ({ ipc, args, cmdArgs, link, storage, detach
     state.update({ config: await ipc.config() })
 
     const pear = new API(ipc, state)
+
     global.Pear = pear
 
+    /* global Pear */
```
davidmarkclements, line 115:
```suggestion
    /* global Pear */
```

PR #197, opened 2024-06-17
test can spawn fixture as a worker

rafapaezbas:


davidmarkclements:
closing - not worth the branching if we need both and we've found this is to do with missing (now incoming) bare functionality for duplex pipe overlapping support

PR #203, opened 2024-06-21
Add HTTP/HTTPS hostname whitelisting for Pear desktop apps

jkcdarunday:
Allows only requests to the sidecar host (127.0.0.1:9342) by default and returns 404 for all other requests. Additional hosts can be whitelisted by listing them as values in the pear.links object in package.json 

Example: adding `{ google: 'https://google.com' }` in `pear.links` will allow all requests to google.com (HTTPS) and all its subpaths (google.com/test etc.)

### DIFF
--- a/subsystems/sidecar/index.js
+++ b/subsystems/sidecar/index.js
@@ -443,6 +443,7 @@ class Sidecar extends ReadyResource {
     }
     if (typeof pkg?.pear?.links === 'object' && pkg.pear.links !== null) {
       for (const link of Object.values(pkg.pear.links)) {
+        if (link.slice(0, 7).toLowerCase() !== 'pear://') continue
         try {
           trusted.add(hypercoreid.encode(hypercoreid.decode(link)))
         } catch {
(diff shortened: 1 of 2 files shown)


subsystems/sidecar/index.js
```diff
@@ -443,6 +443,7 @@ class Sidecar extends ReadyResource {
     }
     if (typeof pkg?.pear?.links === 'object' && pkg.pear.links !== null) {
       for (const link of Object.values(pkg.pear.links)) {
+        if (link.slice(0, 7).toLowerCase() !== 'pear://') continue
```
davidmarkclements, line 446: use parseLink to get proto

jkcdarunday: @davidmarkclements Hmm.. parseLink does not seem to return proto. Did you mean `new URL()`?

davidmarkclements: nice solution with key check, parseLink coming from entrypoints contains proto, but this key check will work either way - nice resolution thank you

PR #238, opened 2024-07-29
Rename soft restart to reload

jkcdarunday:


lib/api.js
```diff
@@ -97,14 +97,21 @@ class API {
...
   restart = async (opts = {}) => {
-    // TODO: use Pear.shutdown when it lands instead
-    if (this.#state.type === 'terminal' && !opts.platform) {
-      Bare.exit(BARE_RESTART_EXIT_CODE)
-    } else {
-      const restart = this.#reftrack(this.#ipc.restart(opts))
-      if (opts.platform && (opts.hard || opts.hard === undefined)) Bare.exit()
-      return restart
+    const restart = this.#reftrack(this.#ipc.restart({ ...opts }))
+    if (opts.platform) Bare.exit()
```
davidmarkclements, line 101: clients shouldn't self-exit they should exit when sidecar exits
```suggestion
```

lib/api.js
```diff
@@ -97,14 +97,20 @@ class API {
...
 
   restart = async (opts = {}) => {
-    // TODO: use Pear.shutdown when it lands instead
-    if (this.#state.type === 'terminal' && !opts.platform) {
-      Bare.exit(BARE_RESTART_EXIT_CODE)
-    } else {
-      const restart = this.#reftrack(this.#ipc.restart(opts))
-      if (opts.platform && (opts.hard || opts.hard === undefined)) Bare.exit()
-      return restart
+    const restart = this.#reftrack(this.#ipc.restart({ ...opts }))
```
davidmarkclements, line 100:
```suggestion
    const restart = this.#reftrack(this.#ipc.restart({ ...opts, hard: true }))
```

lib/api.js
```diff
@@ -97,14 +97,20 @@ class API {
...
+  reload = async (opts = {}) => {
+    if (!opts.platform) {
+      // TODO: use Pear.shutdown when it lands instead
+      if (this.#state.type === 'terminal') Bare.exit(BARE_RESTART_EXIT_CODE)
+      else global.location.reload()
+
+      return
     }
+
+    return this.#reftrack(this.#ipc.restart({ hard: false, ...opts }))
```
davidmarkclements, line 113:
```suggestion
    return this.#reftrack(this.#ipc.restart({ ...opts, hard: false }))
```

PR #249, opened 2024-08-05
Add tests for `pear stage`

jkcdarunday:


test/07-commands.test.js
```diff
@@ -0,0 +1,1436 @@
...
+
+  const testId = Math.floor(Math.random() * 100000)
+
+  const tmp = path.join(fixtures, '.tmp')
+  const targetDir = path.join(tmp, `pear-test-${testId}`)
+  const sourceDrive = new LocalDrive(minimal)
+  const targetDrive = new LocalDrive(targetDir)
+  const mirror = sourceDrive.mirror(targetDrive, { prune: false })
+  // eslint-disable-next-line no-unused-vars
+  for await (const val of mirror) { /* ignore */ }
```
rafapaezbas, line 97: I think we can use mirror.done() instead

PR #262, opened 2024-08-16
Bugfix: add err handler sync when streaming bundle

HDegroote:
Previous behaviour could throw an unhandled exception if the stream threw between creation and its being used in `pipelinePromise`.
Current behaviour first gets the metadata (which is async) and then creates the stream sync.

### DIFF
--- a/subsystems/sidecar/lib/http.js
+++ b/subsystems/sidecar/lib/http.js
@@ -165,8 +165,10 @@ module.exports = class Http extends ReadyResource {
         app.warmup({ protocol, batch })
       }
 
-      const meta = await bundle.getMeta(link.filename)
-      const stream = meta === null ? null : bundle.streamFrom(meta)
+      const meta = await this.entry(link.filename)
+      if (meta === null) return
+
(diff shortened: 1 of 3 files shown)


subsystems/sidecar/lib/http.js
```diff
@@ -164,7 +164,9 @@ module.exports = class Http extends ReadyResource {
         }
         app.warmup({ protocol, batch })
       }
-      const stream = await bundle.streamFrom(link.filename)
+
+      const meta = await bundle.getMeta(link.filename)
+      const stream = meta === null ? null : bundle.streamFrom(meta)
```
HDegroote, line 169: I kept the behaviour exactly as-is, but I'm 99% sure there's another bug here: when meta is null, `      await streamx.pipelinePromise(stream, res)` throws. Unsure what the expected behaviour should be though.

HDegroote:
Includes 2 other changes:
- A bugfix for the case where the entry is null (should not call pipelinePromise then, but early return)
- capture is now a sync method

PR #265, opened 2024-08-20
Sync stream err handler

davidmarkclements:
reopening from #262, see #264 revert - pear run pear://keet fails with this.entry error

### DIFF
--- a/subsystems/sidecar/lib/http.js
+++ b/subsystems/sidecar/lib/http.js
@@ -165,8 +165,8 @@ module.exports = class Http extends ReadyResource {
         app.warmup({ protocol, batch })
       }
 
-      const meta = await this.entry(link.filename)
-      if (meta === null) return
+      const meta = await bundle.entry(link.filename)
+      if (meta === null) throw new ERR_HTTP_NOT_FOUND(`Not Found: "${link.filename}"`)
 
       const stream = bundle.streamFrom(meta)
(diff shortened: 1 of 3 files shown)


subsystems/sidecar/lib/http.js
```diff
@@ -164,7 +164,11 @@ module.exports = class Http extends ReadyResource {
         }
         app.warmup({ protocol, batch })
       }
-      const stream = await bundle.streamFrom(link.filename)
+
+      const meta = await this.entry(link.filename)
```
mafintosh, line 168: bundle.entry

subsystems/sidecar/lib/http.js
```diff
@@ -164,7 +164,11 @@ module.exports = class Http extends ReadyResource {
...
+      if (meta === null) return
```
mafintosh, line 169: this needs to also end the http request with a 404

HDegroote: I did as elsewhere in the fail and now throw a ERR_HTTP_NOT_FOUND error, but there's a bug unrelated to this PR: there's no special case for this error in the catch-handler, so it returns a 500 code instead of a 404: https://github.com/holepunchto/pear/blob/d984e3bd9839e5444fbc3d8a485a1377cdf96d7b/subsystems/sidecar/lib/http.js#L46-L54

PR #269, opened 2024-08-22
return 404 from sidecar on ERR_HTTP_NOT_FOUND

HDegroote:
ERR_HTTP_NOT_FOUND is an expected kind of error, so I think it should return 404 instead of an internal-server error

<strike>Note: I can find no reference to the existing 2 error codes anywhere in the code base, so if those are supposed to be errors defined at Pear level, I believe they are outdated.</strike>
This has been clarified: those errors come from dependencies of pear, not from pear itself


HDegroote:
The CI failure is consistent with other runs:
linux succeeds before all tests ran
mac ran normally and succeeded
windows times out on a test

PR #271, opened 2024-08-22
Fix `--checkout release` flag not working when doing `pear dump`

jkcdarunday:
Fixes bug where the `--checkout release` flag in `pear dump` dumps the most recently staged instead of the last release


jkcdarunday:
Not yet supported, not a bug. Discussed in Slack.

PR #283, opened 2024-08-29
added info for encrypted apps

rafapaezbas:


cmd/index.js
```diff
@@ -106,6 +106,8 @@ module.exports = async (ipc, argv = Bare.argv.slice(1)) => {
     flag('--metadata', 'View metadata only'),
     flag('--key', 'View key only'),
     flag('--json', 'Newline delimited JSON output'),
+    flag('--encryption-key <name>', 'Application encryption key'),
+    flag('--no-ask', 'Suppress permissions dialog'),
```
davidmarkclements, line 110: dialogs

cmd/info.js
```diff
@@ -40,11 +41,20 @@ const output = outputter('info', {
   keys,
   info,
   changelog,
-  error: ({ code, stack }) => `Info Error (code: ${code || 'none'}) ${stack}`
+  error: (err, info, ipc) => {
+    if (err.info && err.info.encrypted && info.ask) {
+      const explain = 'This application is encrypted.\n' +
+        '\nEnter the password to retrieve app info.\n\n'
```
davidmarkclements, line 47:
```suggestion
        '\nEnter the password to retrieve info.\n\n'
```

cmd/info.js
```diff
@@ -40,11 +41,20 @@ const output = outputter('info', {
   keys,
   info,
   changelog,
-  error: ({ code, stack }) => `Info Error (code: ${code || 'none'}) ${stack}`
+  error: (err, info, ipc) => {
```
davidmarkclements, line 44: I know this exists already but I think we should actually throw instead of using error tag

test/07-encrypted.test.js
```diff
@@ -54,11 +54,11 @@ class Rig {
 
 const rig = new Rig()
 
-test('encrypted test setup', rig.setup)
+test.hook('encrypted test setup', rig.setup)
```
davidmarkclements, line 57: nice

cmd/info.js
davidmarkclements: this is used elsewhere also - I think separate to this we should use try catch and not error tag for ops

davidmarkclements: but not change needed here

PR #293, opened 2024-09-03
mask password input

rafapaezbas:
This PR masks the CLI password input


lib/interact.js
```diff
@@ -53,7 +79,8 @@ module.exports = class Interact {
 
   async #input (prompt) {
     stdio.out.write(prompt)
-    const answer = (await once(stdio.in, 'data')).toString()
+    this._rl._prompt = prompt
+    const answer = (await once(this._rl, 'data')).toString()
     return answer.trim() // remove return char
```
davidmarkclements, line 84: this does more than remove /n or /r/n it removes any whitespace round the answer, which I'm fine with but the comment isn't needed

PR #297, opened 2024-09-05
added password dialog for encrypted templates

rafapaezbas:


cmd/init.js
```diff
@@ -39,12 +39,22 @@ module.exports = (ipc) => async function init (cmd) {
...
   } catch (err) {
-    if (err.code !== 'ERR_PERMISSION_REQUIRED') throw err
-    const explain = 'Be sure that software is trusted before using it\n' +
-    '\nType "TRUST" to allow template initialization or anything else to exit\n\n'
-    const ask = 'Trust template'
-    const act = 'Use pear init again to initalize from trusted template'
-    await trust({ ipc, key: err.info.key, message: err.message, explain, ask, act })
+    if (err.code !== 'ERR_PERMISSION_REQUIRED') {
+      throw err
+    } else {
```
davidmarkclements, line 44: if you throw you exit the function - no need for the else branch - also means you can de-nest 
```
if (err.code !== 'ERR_PERMISSION_REQUIRED') throw err
//rest of code at this indent
```

cmd/init.js
```diff
@@ -39,12 +39,22 @@ module.exports = (ipc) => async function init (cmd) {
...
+        const message = 'Added encryption key, run init again to complete it.'
+        await password({ ipc, key: err.info.key, explain, message })
+      } else {
+        const explain = 'Be sure that software is trusted before using it\n' +
+          '\nType "TRUST" to allow template initialization or anything else to exit\n\n'
+        const ask = 'Trust template'
+        const act = 'Use pear init again to initalize from trusted template'
+        await trust({ ipc, key: err.info.key, message: err.message, explain, ask, act })
+      }
+    }
```
davidmarkclements, line 57: this is what iface `permit` should actually do 

this code from 45-57 here should just be one line

```
await permit({ ipc, info: err.info, message: err.message })
```

and then permit should contain the conditional logic to then call `trust` / `password `

as for the current iface `permit`, it's just inline its current function body into `trust` and `password`

init/index.js
```diff
@@ -22,9 +22,8 @@ async function init (link, dir, { ipc, header, autosubmit, defaults, force = fal
   let params = null
   if (isPear) {
     const { drive } = parseLink(link)
-    const trusted = new Set(await ipc.getPreference({ key: 'trusted' }) || [])
-    hypercoreid.encode(drive.key, hypercoreid.encode(drive.key), trusted)
-    if (trusted.has(hypercoreid.encode(drive.key)) === false) {
+    const trusted = await ipc.trusted()
```
davidmarkclements, line 25: let's do this: if (await ipc.trusted(drive.key) === false) throw...

cmd/init.js
rafapaezbas: the issue is that `trust` and `password` all show different messages for stage, dump, run, init and info. Thats why we pass `explain`, `ask` and `act` as arguments.

cmd/init.js
rafapaezbas: updated

cmd/init.js
davidmarkclements: @rafapaezbas right right- so permit({ dialog: 'stage', .... })

and then the permit function owns all the output strings, so it passes the correct explain, ask, act args to trust and password

rafapaezbas: done

davidmarkclements: @rafapaezbas 
> this code from 45-57 here should just be one line

you're still not calling permit directly?

rafapaezbas: done

PR #303, opened 2024-09-10
ci self pear

davidmarkclements:
also introduces

`pear sidecar shutdown`

and modifies `pear info --key` to just print key with no additional visual-framing chars


rafapaezbas:
I think we can remove workflow cache now since seems like its not used anymore https://github.com/holepunchto/pear/blob/main/.github/workflows/ci.yml#L34-L38

PR #304, opened 2024-09-11
fix password dialog on first run of encrypted app

rafapaezbas:


subsystems/sidecar/index.js
```diff
@@ -789,7 +789,7 @@ class Sidecar extends ReadyResource {
       await appBundle.calibrate()
     } catch (error) {
       if (error.code === 'DECODING_ERROR') {
-        const err = new ERR_PERMISSION_REQUIRED('Permission required to run key', state.key)
+        const err = new ERR_PERMISSION_REQUIRED('Permission required to run key', state.key, true)
```
davidmarkclements, line 792: can we make ERR_PERMISSION_REQUIRED so its (msg, info), so we can do `new ERR_PERMISSION_REQUIRED('Permission required to run key', { key: state.key, encrypted: true })`

also for consisitency-sake would you mind making this block be } catch (err) { } and then just `app.report({ err: new ERR_PERMISSION_REQUIRED(...) })`

subsystems/sidecar/ops/stage.js
```diff
@@ -75,8 +75,7 @@ module.exports = class Stage extends Opstream {
     const encryptionKey = encryptionKeys[hypercoreid.normalize(key)] || await secrets.get(params.encryptionKey)
 
     if (encrypted === true && !encryptionKey && !params.encryptionKey) {
-      const err = ERR_PERMISSION_REQUIRED('encryption key required', key, true)
-      throw err
+      throw new ERR_PERMISSION_REQUIRED('encryption key required', { key, encrypted: true })
```
davidmarkclements, line 78:
```suggestion
      throw new ERR_PERMISSION_REQUIRED('Encryption key required', { key, encrypted: true })
```

PR #323, opened 2024-09-25
Add blocking of spindown when platform update is ongoing

jkcdarunday:


test/06-shutdown.test.js
```diff
@@ -0,0 +1,227 @@
...
+test.hook('prepare throttled platform', async (t) => {
+  t.timeout(120_000)
+
+  const patchedArtefactDir = path.join(Helper.tmp, 'arcv-pear')
+  t.comment(`\tCopying platform code to ${patchedArtefactDir}`)
+  await fs.promises.mkdir(patchedArtefactDir, { recursive: true })
+  t.teardown(() => Helper.gc(patchedArtefactDir))
+  const mirror = new LocalDrive(rig.artefactDir).mirror(new LocalDrive(patchedArtefactDir), {
+    prune: false,
+    filter: (key) => !(key.startsWith('/pear/') || key.startsWith('/.git/') || key.startsWith('/test/'))
```
rafapaezbas, line 126: use `ignore` instead of `filter`, will make the test run faster in windows

test/06-shutdown.test.js
```diff
@@ -0,0 +1,227 @@
...
+test.hook('prepare low-spindown platform', async (t) => {
+  t.timeout(120_000)
+
+  const patchedArtefactDir = path.join(Helper.tmp, 'als-pear')
+  t.comment(`Copying platform code to ${patchedArtefactDir}`)
+  await fs.promises.mkdir(patchedArtefactDir, { recursive: true })
+  t.teardown(() => Helper.gc(patchedArtefactDir))
+  const mirror = new LocalDrive(rig.artefactDir).mirror(new LocalDrive(patchedArtefactDir), {
+    prune: false,
+    filter: (key) => !(key.startsWith('/pear/') || key.startsWith('/.git/') || key.startsWith('/test/'))
```
rafapaezbas, line 52: same as below

test/06-shutdown.test.js
```diff
@@ -0,0 +1,227 @@
...
+  const hasSpunDown = await Promise.race([untilExit, timeoutUntil])
+  if (hasSpunDown === false) {
+    t.fail('sidecar failed to spin down')
+  } else {
+    t.pass('sidecar has spun down')
+  }
+})
+
+let platformDirRcv
+test.hook('prepare throttled platform', async (t) => {
```
rafapaezbas, line 117: Do we need to mark this test as a hook?

test/06-shutdown.test.js
```diff
@@ -0,0 +1,227 @@
...
+  const dhtBootstrap = global.Pear.config.dhtBootstrap.map(e => `${e.host}:${e.port}`).join(',')
+  const sidecar = spawn(path.join(platformDirLs, 'current', BY_ARCH), ['--dht-bootstrap', dhtBootstrap, 'sidecar', '--key', staged.key], { stdio: 'pipe' })
+  t.teardown(() => { if (sidecar.exitCode === null) sidecar?.kill() })
+  const untilExit = new Promise(resolve => sidecar.once('exit', resolve))
+  t.teardown(async () => untilExit, { order: Infinity })
+
+  t.comment(`\tWaiting for sidecar spindown timeout to lapse (${(SPINDOWN_TIMEOUT + 10_000) / 1000}s)`)
+  const timeoutUntil = new Promise((resolve) => {
+    untilExit.finally(() => {
+      if (timeout) clearTimeout(timeout)
```
rafapaezbas, line 205: where is timeout defined?

test/06-shutdown.test.js
jkcdarunday: @rafapaezbas Hmm.. I can merge this in the test where it's being used, but one issue is that the execution time of this step is inconsistent (last run was 30s on Linux and 60s on Windows, but I've seen this step go up to 90s). Right now, I have the timeouts for the spindown test a little tight but if I put this with the test, I'll have to loosen the timeout.

Would it be preferable to put this in the test instead?

PR #329, opened 2024-09-28
Persistent DHT cache

AndreiRegiani:
```bash

cat /Users/user/holepunchto/pear/pear/dht.json

{
    "nodes":[
        { "host": "127.0.0.1", "port": 49001 },
        { "host": "127.0.0.2", "port": 49002 }
    ]
}

```

### DIFF
--- a/subsystems/sidecar/index.js
+++ b/subsystems/sidecar/index.js
@@ -32,7 +32,8 @@ const { version } = require('../../package.json')
 const deriveEncryptionKey = require('pw-to-ek')
 const {
   PLATFORM_DIR, PLATFORM_LOCK, SOCKET_PATH, CHECKOUT, APPLINGS_PATH,
-  SWAP, RUNTIME, DESKTOP_RUNTIME, ALIASES, SPINDOWN_TIMEOUT, WAKEUP, SALT
+  SWAP, RUNTIME, DESKTOP_RUNTIME, ALIASES, SPINDOWN_TIMEOUT, WAKEUP,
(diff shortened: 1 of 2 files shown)


mafintosh:
Store should be named dht instead of known-nodes and the key you set simply nodes instead of dht

subsystems/sidecar/index.js
```diff
@@ -892,6 +894,20 @@ class Sidecar extends ReadyResource {
...
   }
 
+  #setKnownNodes (dht) {
+    if (dht) knownNodes.set('dht', dht)
+  }
+
+  #getKnownNodes () {
+    let dht = this.swarm.dht.toArray({ limit: KNOWN_NODES_LIMIT })
+    if (dht) {
+      this.#setKnownNodes(dht)
```
mafintosh, line 904: Dont think it should set them here, just on teardown explicitly

subsystems/sidecar/index.js
```diff
@@ -892,6 +894,20 @@ class Sidecar extends ReadyResource {
     yield * preferences.entries()
   }
 
+  #setKnownNodes (dht) {
+    if (dht) knownNodes.set('dht', dht)
```
davidmarkclements, line 898: avoid redirection without purpose - just set inline where you need it

subsystems/sidecar/index.js
davidmarkclements: so function isn't needed then, just do this all inline in shutdown

mafintosh (approved):

PR #336, opened 2024-10-03
move to drive-bundler v2

mafintosh:


mafintosh:
depends on https://github.com/holepunchto/pear-updater/pull/19 also

PR #339, opened 2024-10-07
[NEW] Persistent DHT cache

AndreiRegiani:
```bash

cat /Users/user/holepunchto/pear/pear/dht.json

{
    "nodes":[
        { "host": "127.0.0.1", "port": 49001 },
        { "host": "127.0.0.2", "port": 49002 }
    ]
}

```

### DIFF
--- a/subsystems/sidecar/index.js
+++ b/subsystems/sidecar/index.js
@@ -961,7 +961,7 @@ class Sidecar extends ReadyResource {
     if (this.swarm) {
       if (!this.dhtBootstrap) {
         const nodes = this.swarm.dht.toArray({ limit: KNOWN_NODES_LIMIT })
-        if (nodes.length) knownNodes.set('nodes', nodes)
+        if (nodes.length) await knownNodes.set('nodes', nodes)
       }
       await this.swarm.destroy()
     }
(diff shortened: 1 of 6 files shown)


subsystems/sidecar/index.js
```diff
@@ -697,7 +699,7 @@ class Sidecar extends ReadyResource {
     const id = client.userData?.id || `${client.id}@${startId}`
     const app = client.userData = client.userData?.id ? client.userData : new this.App({ id, startId, session })
     await this.ready()
-    const dht = this.swarm.dht.toArray({ limit: 20 })
+    const dht = { nodes: this.swarm.dht.toArray({ limit: KNOWN_NODES_LIMIT }) || [] }
```
rafapaezbas, line 702: I think it should be: `const dht = { nodes: this.swarm.dht.toArray({ limit: KNOWN_NODES_LIMIT }) || [], bootstrap: this.dhtBootstrap }`

mafintosh: in the same spirit, very important!, if custom bootstraps are set DO NOT persist or load the node cache from disk. will create intense debugging scenarios

rafapaezbas (approved):

PR #341, opened 2024-10-07
mac update restart fix for ui apps

davidmarkclements:


davidmarkclements:
(just needed `pear gc releases`)

PR #349, opened 2024-10-10
[GHA] bump-deps and bump-versions

maidh91:
# Pre-PRs
https://github.com/holepunchto/actions/pull/1

# Summary
- Add GHA to bump deps and bump versions
- Then it will create PR

### DIFF
--- a/.github/workflows/bump-deps.yml
+++ b/.github/workflows/bump-deps.yml
@@ -1,0 +1,9 @@
+name: Bump Deps
+
+on:
+  workflow_dispatch:
+
+jobs:
+  bump-deps:
+    # TODO: Use @main after merging https://github.com/holepunchto/actions/pull/1
+    uses: holepunchto/actions/.github/workflows/bump-deps-v2.yml@87f1c8e3ef71f8659b08a4dbed0e28b3068aad3f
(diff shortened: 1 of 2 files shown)


scripts/build.sh
```diff
@@ -0,0 +1,39 @@
...
+
+  rm -fr by-arch
+  npm run archdump
+
+  rm -rf ./node_modules
+  npm i --omit=dev
+  npm run prune
+
+  ./pear.dev sidecar &
+  ./pear.dev seed dev &
```
mafintosh, line 24: not forking is allowed in the build script

scripts/build.sh
```diff
@@ -0,0 +1,39 @@
...
+  rm package-lock.json
+  ./pear.dev stage dev
+fi
+
+if [ $channel == "staging" ]; then
+  pear seed staging &
```
mafintosh, line 30: same as above

maidh91: removed

scripts/build.sh
maidh91: removed

package.json
```diff
@@ -9,6 +9,7 @@
     "lint": "standard",
     "lint:fix": "standard --fix",
     "archdump": "node scripts/bootstrap.js --archdump",
+    "prestage": "npm run archdump",
```
rafapaezbas, line 12: I think we need to run prune-dual-prebuilds.mjs for now

PR #356, opened 2024-10-12
Extend the handler of the option --unsafe-clear-preferences to clear both preferences and permits

maidh91:
# Context
- We would like doctor to test TrustDialog for new key
- To do that, we need to clear all `trusted` keys
- Then click/run on the key again
- The TrustDialog should open

# Summary
- extend the handler of the option `--unsafe-clear-preferences` to clear both preferences and permits

# Test
```
./pear.dev run --unsafe-clear-preferences pear://runtime
./pear.dev run pear://kpzdysqu3uqu3nz7xakidqb1ju71jbzaszxu58r3owodbujyghuy
```
image

### DIFF
--- a/subsystems/sidecar/index.js
+++ b/subsystems/sidecar/index.js
@@ -467,6 +467,13 @@ class Sidecar extends ReadyResource {
     return permits.set('trusted', Array.from(trusted))
   }
 
+  async untrust (key) {
+    const z32 = hypercoreid.encode(key)
+    const trusted = new Set((await permits.get('trusted')) || [])
+    trusted.splice(trusted.indexOf(z32), 1); // remove key from trusted
+    return permits.set('trusted', Array.from(trusted))
+  }
+
   async trusted (key) {
     const z32 = hypercoreid.encode(key)
     const trusted = await permits.get('trusted') || []
--- a/subsystems/sidecar/index.js
+++ b/subsystems/sidecar/index.js
@@ -470,7 +470,7 @@ class Sidecar extends ReadyResource {
   async untrust (key) {
     const z32 = hypercoreid.encode(key)
     const trusted = new Set((await permits.get('trusted')) || [])
-    trusted.splice(trusted.indexOf(z32), 1); // remove key from trusted
+    trusted.splice(trusted.indexOf(z32), 1) // remove key from trusted
     return permits.set('trusted', Array.from(trusted))
   }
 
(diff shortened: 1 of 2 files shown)


subsystems/sidecar/index.js
```diff
@@ -467,6 +467,13 @@ class Sidecar extends ReadyResource {
     return permits.set('trusted', Array.from(trusted))
   }
 
+  async untrust (key) {
+    const z32 = hypercoreid.encode(key)
+    const trusted = new Set((await permits.get('trusted')) || [])
+    trusted.splice(trusted.indexOf(z32), 1); // remove key from trusted
```
AndreiRegiani, line 473: JS `standard` style doesn't use semicolon. Try `npm run lint:fix`

maidh91: fixed, thanks

rafapaezbas:
awesome, can you add a test for this? You can follow the smoke test with the next additional steps: 
-  Untrust the key after staging
-  Run the app and verify that the app is untrusted
-  Trust the app
-  Run the app and verify that is trusted

subsystems/sidecar/index.js
```diff
@@ -467,6 +467,13 @@ class Sidecar extends ReadyResource {
     return permits.set('trusted', Array.from(trusted))
   }
 
+  async untrust (key) {
```
HDegroote, line 470: Not specific to this PR, but I believe the `permit` and this `untrust` method are susceptible to race conditions. If permit/untrust runs overlap, entries can be lost, because of the async `permits.get` and `permits.set` methods.

Note: I'm not 100% sure they can overlap in practice, but I don't see any debouncing or other race-condition handling going on here, so it seems like they can.

For `permit` this is annoying, because a user might have to `permit` again. But for `untrust` this is potentially dangerous, because a user will believe he has 'untrusted' an app, while in reality it is still trusted.

maidh91: good point, perhaps to simplify it, we can add `clearPermits` to remove all permits instead of untrust a single one?

maidh91: `await permits.clear()`

maidh91: removed this fn

maidh91: fyi, there is a task to fix the race condition

https://holepunchdev.slack.com/archives/C054FE134P6/p1728905127730489?thread_ts=1728893299.926549&cid=C054FE134P6

maidh91:
closed due to the fix from @rafapaezbas 

https://github.com/holepunchto/pear/pull/373#pullrequestreview-2372481180

PR #357, opened 2024-10-14
fixed timeout teardown for electron context

rafapaezbas:
This PR fixes timeout teardown for desktop apps. It adds a check for global Bare, in case it does not exist, uses `process.exit` in order to close the app.

### DIFF
--- a/gui/gui.js
+++ b/gui/gui.js
@@ -477,6 +477,10 @@ class App {
       }
     })
 
+    electron.app.on('app-exit', () => {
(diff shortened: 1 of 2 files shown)


mafintosh:
process.exit in electron does what we want here? ie it should still graceful exit the electron app

mafintosh:
(i'm no expert but thought you had to do app.exit for that in electron)

mafintosh (approved):

PR #399, opened 2024-10-28
HyperDB initial integration

AndreiRegiani:
`./pear.dev sidecar --log-level TRACE`

image


subsystems/sidecar/index.js
```diff
@@ -934,7 +935,8 @@ class Sidecar extends ReadyResource {
     }
     this.keyPair = await this.corestore.createKeyPair('holepunch')
     if (this.dhtBootstrap) LOG.info('sidecar', 'DHT bootstrap set', this.dhtBootstrap)
-    const nodes = this.dhtBootstrap ? undefined : await knownNodes.get('nodes')
+    const knownNodes = (await db.findOne('@pear/dht'))?.nodes
```
mafintosh, line 938: db.get since its not an index, much faster

subsystems/sidecar/index.js
```diff
@@ -36,9 +37,9 @@ const {
   SALT, KNOWN_NODES_LIMIT
 } = require('../../constants')
 const { ERR_INTERNAL_ERROR, ERR_PERMISSION_REQUIRED } = require('../../errors')
+const db = HyperDB.rocks(path.join(PLATFORM_DIR, 'hyperdb'), require('../../spec/db'))
```
mafintosh, line 40: we should prop move the path into constants, ie PLATFORM_HYPERDB

davidmarkclements, line 40: const definition = require('../spec/db') 

and then pass it here 

also does `./hyperdb` need to be added to .gitignore?

PR #403, opened 2024-10-30
use Pear.worker for smoke test  instead of spawn process

maidh91:
new one: https://github.com/holepunchto/pear/pull/404


test/01-smoke.test.js
```diff
@@ -38,17 +38,25 @@ test('smoke', async function ({ ok, is, plan, comment, teardown, timeout, end })
...
+    } else if (obj.type === 'dhtBootstrap') {
+      is(JSON.stringify(obj.value), JSON.stringify(Pear.config.dht.bootstrap), 'dht bootstrap matches Pear.config.dth.bootstrap')
+    }
+  })
+  pipe.on('end', () => {
+    ok(true, 'pipe ended')
+  })
+
+  pipe.write('versions')
+  await new Promise((resolve) => setTimeout(resolve, 1000))
```
rafapaezbas, line 56: I think we should await pipe response instead of using `setTimeout`.

Something like:

```
async request (pipe, msg) => new Promise((resolve) => {
  pipe.on('data', resolve)
  pipe.write(msg)
}

request(worker, 'version')
```

test/01-smoke.test.js
```diff
@@ -3,16 +3,16 @@ const test = require('brittle')
 const path = require('bare-path')
 const hypercoreid = require('hypercore-id-encoding')
 const Helper = require('./helper')
-const harness = path.join(Helper.localDir, 'test', 'fixtures', 'harness')
+const worker = path.join(Helper.localDir, 'test', 'fixtures', 'worker')
```
rafapaezbas, line 6: Lets create another fixture, we can call it `basic worker`, the plan is to replace harness with a much simpler one

test/03-worker.test.js
```diff
@@ -17,7 +17,7 @@ test('worker pipe', async function ({ is, plan, comment, teardown }) {
   const messages = []
   const response = new Promise((resolve) => {
     pipe.on('data', (data) => {
-      messages.push(data.toString())
+      messages.push(JSON.parse(data.toString()).value)
```
rafapaezbas, line 20: As mentioned above, lets create new fixture so we dont have to change this test yet.

maidh91: updated

test/01-smoke.test.js
maidh91: updated

test/01-smoke.test.js
```diff
@@ -38,17 +39,31 @@ test('smoke', async function ({ ok, is, plan, comment, teardown, timeout, end })
...
+      is(JSON.stringify(obj.value), JSON.stringify(Pear.config.dht.bootstrap), 'dht bootstrap matches Pear.config.dth.bootstrap')
+      dhtBootstrapPromise.resolve()
+    }
+  })
+  pipe.on('end', () => {
+    ok(true, 'pipe ended')
+  })
+
+  pipe.write('versions')
+  await versionsPromise.promise
```
rafapaezbas, line 63: sorry I think I miss explained, with my comment to have a test like:
```
const version = await workerRequest(pipe, 'version')
is(version, expected)
```

That makes the test more readable imo

test/fixtures/basic-worker/index.js
```diff
@@ -0,0 +1,20 @@
+const pipe = Pear.worker.pipe()
+
+pipe.on('data', async (data) => {
+  const type = data.toString()
```
rafapaezbas, line 4: we can simplify this, we dont need to include the type in the response because test requests are going to be awaited, so there is no need to check the type from the test

test/test.util.js
```diff
@@ -0,0 +1,18 @@
+'use strict'
+function promiseToComplete () {
```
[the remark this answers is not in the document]
maidh91: removed

test/01-smoke.test.js
maidh91: updated

test/fixtures/basic-worker/index.js
maidh91: updated

maidh91: updated

maidh91:
move to https://github.com/holepunchto/pear/pull/404

PR #406, opened 2024-10-31
Move version flag to main `pear` command

lejeunerenard:
This PR was motivated by the `-v` and the `--sidecar` being listend in the output from `pear help sidecar`. The verbose flag (`-v`) is only used [in the main `pear` command](https://github.com/holepunchto/pear/blob/9f9283fb291e45b1b7b2c1dee1e276832e3ade84/cmd/index.js#L217) that I can tell.

I tried moving `--sidecar` as well since the `sidecar` command doesn't seem to use it (and it sounds redundant to 'Boot Sidecar' from a command to run a sidecar in the terminal), but when I did this tests no longer passed. I tracked it down to the flag being added via [`shell.js`](https://github.com/holepunchto/pear/blob/9f9283fb291e45b1b7b2c1dee1e276832e3ade84/shell.js#L4) and triggered by [`lib/tryboot.js`](https://github.com/holepunchto/pear/blob/9f9283fb291e45b1b7b2c1dee1e276832e3ade84/lib/tryboot.js#L27). So I left it in for the moment, but it'd be nice to not include it in `pear sidecar` if it indeed doesn't do anything.

### DIFF
--- a/cmd/index.js
+++ b/cmd/index.js
@@ -208,6 +208,7 @@ module.exports = async (ipc, argv = Bare.argv.slice(1)) => {
     encryptionKey,
     versions,
     help,
+    flag('-v', 'Print version'),
     footer(usage.footer),
     bail(explain),
     pear
--- a/def/pear.js
+++ b/def/pear.js
@@ -2,7 +2,6 @@
 const { flag } = require('paparam')
 
 module.exports = [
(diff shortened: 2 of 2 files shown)


rafapaezbas:
fixed in https://github.com/holepunchto/pear/pull/445

PR #431, opened 2024-11-08
add logger

maidh91:


davidmarkclements:
@maidh91 I'm closing this as it's not the direction we need for logging, we can scope down pieces of this maybe but avoid adding complexity at all costs please

PR #434, opened 2024-11-08
Use worker instead of inspector on updates tests

jkcdarunday:


test/05-updates.test.js
```diff
@@ -285,54 +289,53 @@ test('Pear.updates should notify Platform stage, Platform release updates (diffe
...
   const pearUpdate2Length = update2Version.length
 
   is(pearUpdate2Key, rig.key, 'platform release update matches staging key')
   ok(pearUpdate2Length > pearUpdateLength, `platform version length incremented (v${update2Version?.fork}.${update2Version?.length})`)
 
-  await running.inspector.evaluate('__PEAR_TEST__.close()')
-  await running.inspector.close()
-  const { code } = await running.until.exit
-  is(code, 0, 'exit code is 0')
+  pipe.write('exit')
```
maidh91, line 315: can we try `pipe.end()` for parent?

then in the child worker, we can `pipe.on('end', cb)` or `pipe.on('close', cb)` with a `cb` to updates.end()

test/05-updates.test.js
```diff
@@ -27,57 +27,61 @@ test('Pear.updates(listener) should notify when restaging and releasing applicat
...
+  const update2Version = update2?.version
   is(hypercoreid.normalize(update2Version?.key), hypercoreid.normalize(key), 'app updated with matching key')
   is(update2Version?.fork, 0, 'app version.fork is 0')
   ok(update2Version?.length > update1Version?.length, `app version.length incremented (v${update2Version?.fork}.${update2Version?.length})`)
-  await running.inspector.evaluate('__PEAR_TEST__.close()')
-  await running.inspector.close()
-  const { code } = await running.until.exit
-  is(code, 0, 'exit code is 0')
+
+  pipe.write('exit')
```
maidh91, line 83: can we try pipe.end() for parent?

then in the child worker, we can pipe.on('end', cb) or pipe.on('close', cb) with a cb to updates.end()

jkcdarunday: @maidh91 Hmm.. tried this but it didn't emit the end nor the close event on the worker.

maidh91 (approved):

test/05-updates.test.js
jkcdarunday: @rafapaezbas Hmm.. that one seems to be using Pear.teardown() instead of a pipe.on('end', ...) handler on the worker side.

Pear.teardown() won't trigger in the updates worker because Pear.updates() is still listening.

jkcdarunday: Turns out we need to resume() the pipe on the worker side for end to be emitted correctly to switch its stream to flowing mode. Discussed in slack.


https://nodejs.org/api/stream.html#event-end
> The 'end' event will not be emitted unless the data is completely consumed. This can be accomplished by switching the stream into flowing mode, or by calling [stream.read()](https://nodejs.org/api/stream.html#readablereadsize) repeatedly until all data has been consumed.

scripts/test.mjs
```diff
@@ -14,7 +14,6 @@ const root = path.dirname(dirname)
 const force = Bare.argv.includes('--force-install')
 
 const dirs = [
-  path.join(root, 'test', 'fixtures', 'harness', 'node_modules'),
   path.join(root, 'test', 'fixtures', 'encrypted', 'node_modules'),
```
maidh91, line 17: can we also remove this line to npm install for encrypted since encrypted has no deps now

maidh91 (approved):

test/05-updates.test.js
maidh91: great finding, good to know that, thanks @jkcdarunday

PR #478, opened 2024-12-04
improve worker desktop

maidh91:
# Issue
- Run `./pear.dev run -d ./examples/desktop`
- Then in browser console, run `Pear.worker.run('pear://doctor')`
- Now, 
  - case 1: if `Ctrl C` => both desktop apps closed
  - case 2: if we manual close examples/desktop app, doctor does not close

# Summary
This PR handles the case 2 to close all child workers when the parent closes


davidmarkclements:
dead

PR #491, opened 2024-12-11
improve-crash-output

maidh91:
# Default

image

# With `--log` param

image

### DIFF
--- a/lib/crasher.js
+++ b/lib/crasher.js
@@ -26,6 +26,9 @@ function logCrash (logPath, errorInfo, checkout, stackTrace, err) {
 }
 
 function printCrash (errorInfo, stackTrace, err) {
+  if (hasLogged) return // Safety: only logging 1 crash per run
+  hasLogged = true
+
   const errInfo = err !== null && typeof err === 'object' ? JSON.stringify(err, 0, 4).slice(1, -2) : ''
   const errorMsg = `${stackTrace + errInfo}\n\n${errorInfo}`
 
--- a/lib/crasher.js
+++ b/lib/crasher.js
@@ -4,15 +4,15 @@ const fs = isBare ? require('bare-fs') : require('fs')
 const path = isBare ? require('bare-path') : require('path')
 const checkout = require('../checkout')
 const os = isBare ? require('bare-os') : require('os')
+const pid = isBare ? Bare.pid : global.process.pid
 
 let hasLogged = false
 const start = Date.now()
-function logCrashAndExit (logPath, errorInfo, checkout, stackTrace, err) {
+function logCrash (logPath, errorInfo, checkout, stackTrace, err) {
   if (hasLogged) return // Safety: only logging 1 crash per run
   hasLogged = true
 
   const timeStamp = (new Date()).toISOString()
-  const pid = isBare ? Bare.pid : global.process.pid
   const uptime = (Date.now() - start) / 1000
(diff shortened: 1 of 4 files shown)


lib/crasher.js
```diff
@@ -33,19 +33,35 @@ function logCrashAndExit (logPath, errorInfo, checkout, stackTrace, err) {
...
+function printAndExit (errorInfo, stackTrace, err) {
+  // Electron does not play well with process.exit, so use app.exit instead
+  const runContext = isBare ? global.Bare : (global.process.versions.electron ? require('electron').app : global.process)
+  const pid = isBare ? Bare.pid : global.process.pid
+  const errInfo = err !== null && typeof err === 'object' ? JSON.stringify(err, 0, 4).slice(1, -2) : ''
+  const errorMsg = `${stackTrace + errInfo}\n\n${errorInfo}`
+
+  console.error(errorMsg)
+
+  if (isBare) {
```
rafapaezbas, line 45: imo, this part should happen outside the function so we dont have to duplicate

lib/crasher.js
```diff
@@ -33,19 +33,35 @@ function logCrashAndExit (logPath, errorInfo, checkout, stackTrace, err) {
...
+
+function setupCrashHandlers (processName, swap, enableLog) {
   const crashlogPath = path.join(swap, `${processName}.crash.log`)
   const runContext = isBare ? global.Bare : global.process
 
   runContext.on('unhandledRejection', (reason) => {
     const stack = reason?.stack || reason || ''
     const errorInfo = `${processName} exiting due to unhandled rejection`
-    logCrashAndExit(crashlogPath, errorInfo, checkout, stack, reason)
+    enableLog ? logCrashAndExit(crashlogPath, errorInfo, checkout, stack, reason) : printAndExit(errorInfo, stack, reason)
```
rafapaezbas, line 59: I think we should follow the code style convention and use if-else instead of ?

lib/crasher.js
```diff
@@ -33,19 +33,35 @@ function logCrashAndExit (logPath, errorInfo, checkout, stackTrace, err) {
   }
 }
 
-function setupCrashHandlers (processName, swap) {
+function printAndExit (errorInfo, stackTrace, err) {
+  // Electron does not play well with process.exit, so use app.exit instead
+  const runContext = isBare ? global.Bare : (global.process.versions.electron ? require('electron').app : global.process)
```
rafapaezbas, line 38: some of this definitions are the same as logCrashAndExit, can we dedup?

lib/crasher.js
maidh91: updated

lib/crasher.js
maidh91: updated

lib/crasher.js
```diff
@@ -25,27 +23,43 @@ function logCrashAndExit (logPath, errorInfo, checkout, stackTrace, err) {
...
+  if (enableLog) {
+    logCrash(logPath, errorInfo, checkout, stack, err)
+  } else {
+    printCrash(errorInfo, stack, err)
+  }
 
   if (isBare) {
     os.kill(pid)
   } else {
+    const runContext = isBare ? global.Bare : (global.process.versions.electron ? require('electron').app : global.process)
```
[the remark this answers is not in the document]
maidh91: updated

lib/crasher.js
maidh91: updated

PR #514, opened 2024-12-27
Refactor until result

rafapaezbas:


maidh91 (approved):
LGTM

would be great if we can add some comment, or set `const STOP_CHAR = `\n`` at the top of the `untilResult` fn to quickly remind of the ending condition

PR #520, opened 2025-01-03
log-not-allowed-links

maidh91:


davidmarkclements:
closing - agree that messaging should improve but should be http and we'll do it after v2

PR #542, opened 2025-01-14
Rename "hideOnClose" to "hideable" and apply option on interload

educastellano:


rafapaezbas:
Looks good to me, tested on Linux and hides the window even when running from pear://[app-key]

PR #561, opened 2025-01-23
Add Pear.tray() for desktop apps

maidh91:
# API
```
Pear.tray(opt, listener)

opts: {
  icon: string,
  menu: { [key]: label },
  os: { win32: bool, linux: bool, darwin: bool }
}

listener: {
  [menuKey]: handler
}
```

# Default values
- `icon`: Pear icon
- `menu`: Show and Quit
  ```
  {
    show: 'Show',
    quit: 'Quit'
  }
  ```
- `os`: enabled for all OS
  ```
  { win32: true, linux: true, darwin: true }
  ``` 

# Examples
`Pear.tray()`
image

`Pear.tray({ icon: 'icon.png' })`
image

```
Pear.tray(
  { icon: 'icon.png', menu: { foo: 'Foo', bar: 'Bar' } }, 
  console.log
)
```
image

# Videos
## Linux
https://github.com/user-attachments/assets/d88c43c1-80d8-4013-b976-b636218f5fd9

## Mac
https://github.com/user-attachments/assets/871c5c62-9d32-4ed3-8b7f-6981cfa6217d

## Windows
https://github.com/user-attachments/assets/8ffccc88-d67c-40c5-ac4f-e976de527f93

## Custom icon
image


gui/gui.js
```diff
@@ -621,6 +621,32 @@ class App {
         }
       })
 
+      const configGuiOptions = state.config.options.gui || state.config.options
+      if (configGuiOptions.hideable || configGuiOptions[process.platform]?.hideable) {
```
educastellano, line 625: maybe handle all these gui options in a single place? we have `guiOptions` above that does already the platform check

maidh91: @educastellano @rafapaezbas 

in local dev, it's ok, but if we stage + seed, and pear run, then `unfilteredGuiOptions` and `guiOptions` are empty, only `state.config.options` has value

image

maidh91: state.options => empty
state.config.options => has value

rafapaezbas:
see refactor proposal to do as @educastellano says https://github.com/holepunchto/pear/pull/565

PR #593, opened 2025-02-04
improve tray darkMode ipc

maidh91:
- ipc only between gui and preload
- remove redundant ipc to sidecar


davidmarkclements:
dead - (if still relevant, please PR pear-electron)

PR #604, opened 2025-02-06
fixed checkout.js requires

rafapaezbas:


mafintosh:
Context is that we have a file on disk called checkout as well so it will try to load that one if running without a resolution map

PR #614, opened 2025-02-13
teardown worker on exit

rafapaezbas:


lib/worker.js
```diff
@@ -72,6 +73,11 @@ class Worker {
     }
     const pipe = new Pipe(fd)
     pipe.on('end', () => {
+      if (isBare) {
+        if (!Bare.exiting) signals.send('SIGTERM', Bare.pid)
```
maidh91, line 77: wonder if it's better to call these on `pipe.on('close', () => {...})`

davidmarkclements:
dead

PR #621, opened 2025-02-20
improve tray default listener

maidh91:
- in Windows, left click will toggle the app to open/hide 
- add `await` on all ipc calls, e.g. `show` should be done before running `focus`


davidmarkclements:
marking defaults for removal after next, consider as dead code until then

PR #699, opened 2025-04-14
[port] set externalCorestore to `$HOME/.pear-archdump/<drive-key>`

maidh91:
https://github.com/holepunchto/pear/pull/698 is merged in main


davidmarkclements:
aligned on next

PR #701, opened 2025-04-16
pear stage --purge & config pear.stage.purge & ignore features

geordangesink:
- new feature: purge: boolean (flag or config) that removes all ignored files from stage
- ignores add onto each other instead of replacing (default, config, flag)
- unignore with ! prefix
- ignore and unignore work with globs (/**/\*someting.\*)


subsystems/sidecar/ops/stage.js
```diff
@@ -145,3 +161,83 @@ module.exports = class Stage extends Opstream {
...
+          if (isNegated) unignores.add(key)
+          else ignores.add(key)
+        }
+      }
+    }
+    this.ignores = ignores
+    this.unignores = unignores
+  }
+
+  getIgnorer () {
```
davidmarkclements, line 225:
```suggestion
  ignorer () {
```

subsystems/sidecar/ops/stage.js
```diff
@@ -71,8 +75,10 @@ module.exports = class Stage extends Opstream {
     const select = only
       ? (key) => only.some((path) => key.startsWith(path[0] === '/' ? path : '/' + path))
       : null
+    const glob = new GlobDrive(src, ignore)
+    await glob.ready()
 
-    const opts = { ignore, dryRun, batch: true, filter: select }
+    const opts = { ignore: glob.getIgnorer(), dryRun, batch: true, filter: select }
```
davidmarkclements, line 81:
```suggestion
    const opts = { ignore: glob.ignorer(), dryRun, batch: true, filter: select }
```

PR #709, opened 2025-04-28
fix: set exit code to 1 during cmd bail on arbitrary error

RamazTs:
Pear process currently returns exit code 0 when detecting a known error. The intention behind this PR is to make the Pear process exit with code 1, regardless of whether the error is known or not.

### DIFF
--- a/cmd/index.js
+++ b/cmd/index.js
@@ -279,5 +279,6 @@ module.exports = async (ipc, argv = Bare.argv.slice(1)) => {
 
     print(reason, false)
     print('\n' + bail.command.usage())
+    Bare.exit(1)
   }
 }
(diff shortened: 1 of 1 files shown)


cmd/index.js
```diff
@@ -279,5 +279,6 @@ module.exports = async (ipc, argv = Bare.argv.slice(1)) => {
 
     print(reason, false)
     print('\n' + bail.command.usage())
+    Bare.exit(1)
```
kasperisager, line 282: Similar to `process.exit()` in Node.js, `Bare.exit()` will terminate execution immediately which may or may not be what we want. Alternatively, `Bare.exitCode = 1` can be used to set the exit code without forcing the exit.

RamazTs: Thanks for the feedback! changed it to `Bare.exitCode = 1`

PR #716, opened 2025-05-01
[pear-next] examples/desktop: closeHides: false

AndreiRegiani:
```
"closeHides": false
```

(same effect as omitting, but serves as documentation for example app)


davidmarkclements:
we should just document - (and/or create an example/template for this)

PR #745, opened 2025-06-06
[pear-next] Fix paparam syntax for `ms` arg in `pear gc interfaces --age <ms>`

lejeunerenard:
`ms` was not correctly parsed when not defined as required or optional in `paparam`'s syntax.


davidmarkclements:
nice catch, no longer a flag

PR #928, opened 2025-11-08
Add tests for preflight flag

Didericis:
- assert that `run` without the `--preflight` flag both downloads assets and runs app
- assert that `run` with the `--preflight` flag only downloads assets


Didericis:
@davidmarkclements should be ready to review, but a test will be failing until PR for [pear-run](https://github.com/holepunchto/pear-run/pull/12) gets merged in, so keeping as draft until that happens

test/12-assets.test.js
```diff
@@ -62,7 +62,8 @@ test('staged manifest assets fetched by run', async (t) => {
     Pear.app.dir = base
   })
   const run = await Helper.run({ link: appWithAssetsDir })
-  await Helper.untilResult(run.pipe)
+  const result = await Helper.untilResult(run.pipe)
+  t.is(result, 'hello')
```
Didericis: Am explicitly checking this now to make sure that running without the `--preflight` tag here behaves differently than running with `--preflight`

test/14-preflight.test.js
```diff
@@ -0,0 +1,100 @@
...
+    flags: ['--preflight']
+  })
+  try {
+    console.log('A')
+    await Helper.untilBail(run.pipe)
+    console.log('B')
+    t.pass()
+  } catch (e) {
+    t.fail(e.message)
+  }
```
geordangesink: you can use [`t.execution`](https://github.com/holepunchto/brittle)

PR #969, opened 2025-12-15
resolve --store relative path

maidh91:


davidmarkclements:
https://github.com/holepunchto/pear-state

PR #1001, opened 2026-01-30
run pear.main field (failing test)

davidmarkclements:


subsystems/sidecar/ops/run.js
```diff
@@ -219,7 +219,10 @@ module.exports = class Run extends Opstream {
 
       LOG.info(LOG_RUN_LINK, id, 'assets', state.assets)
       if (flags.preflight) return { bail: { code: 'PREFLIGHT' } }
-      const bundle = await app.pod.bundle(state.entrypoint)
+      const bundle = await app.pod.bundle(
+        // FIXME: temporary demonstration of fix, should be removed
+        flags.pre ? state.entrypoint : path.join(state.entrypoint, state.main)
```
Didericis: logic that sets `state.entrypoint` should be doing this join, I think

PR #1074, opened 2026-05-04
Switch to new pear-runtime Architecture

geordangesink:
## Pear Runtime + Bare Build
- Added `bare-build` flow for `pear` standalone builds
- Added `pear-runtime` flow for `pear` standalone updates

## Examples
- Removed both `desktop` and `terminal` examples

## CLI + Commands
- Removed `pear run` command from CLI and dispatch.
- Added single `pear run` removal error pointing to `pear-runtime`.
- Removed run-related tests.
- Removed `pear init` + template flow.
- Removed `shift` command and run-era app data surfaces.
- Removed `drop` command.

## Assets + GC
- Removed run/preflight assets pipeline and related tests.
- Removed `gc assets`.

## Stage + Distribution
- Removed `builtins` from stage flow.
- Removed compact stage mode (`pear-shake`) and compact CLI/test expectations.
- Removed warmup pipeline (`drive-analyzer`) end-to-end:
  - no warmup generation in stage,
  - no warmup propagation in provision,
  - no warmup prefetch consumption in pod.
- Stage now defaults to broad mirroring (no tree-shake, no warmup metadata).

## Sidecar + DB
- Replaced `pear-hyperdb` wrapper usage with in-repo spec/hyperdb model path.
- Cleaned schema/db artifacts for removed run-era surfaces (`manifest/current/checkout` migration cleanup items).
- Removed `pear-gunk` runtime plumbing from sidecar bootstrap.

## Runtime API Surface
- Removed `pear-api` dependency.
- Removed operational `global Pear` usage from command/runtime flow.
- Kept minimal RTI bootstrap shim required by current `pear-rti` / `pear-constants` contract.

# TODO

- [x] Linux update flow fix (possible bare-pipe issue)
- [x] remove `pear sidecar --key link` command
- [x] make runtime updater work
- [x] adjust runtime updater (right now it applies as soon as update hits, which kills the sidecar instantly)
- [x] cut away model.js (The new db has these collections: dht, multisig, all others aren't needed after pear run is removed.)
- [x] add Windows builds (x64 and arm64)


geordangesink:
running sidecar command in one shell and then touch in another on windows gives this:

`Uncaught (in promise) FileError: EBUSY: resource busy or locked, open "\\\\?\\C:\\Users\\Geordan\\AppData\\Local\\Temp\\pear.exe-6b284e7e8d871a66f957486121bb627257c100641229989edb7f138792438755\\node_modules\\bare-crypto\\prebuilds\\win32-x64\\bare-crypto.bare"
    at open (bare:/runtime.bundle/node_modules/bare-fs/index.js:136:11)
    at async Object.writeFile (bare:/runtime.bundle/node_modules/bare-fs/index.js:2067:10)
    at async bare:/runtime.bundle/lib/runtime.js:29:7
    at async process (bare:/runtime.bundle/node_modules/bare-unpack/index.js:67:27)
    at async Promise.all (index 56)
    at async unpack (bare:/runtime.bundle/node_modules/bare-unpack/index.js:43:3)
    at async Object.start (bare:/runtime.bundle/lib/runtime.js:18:20) {
  code: 'EBUSY',
  operation: 'open',
  path: '\\\\?\\C:\\Users\\Geordan\\AppData\\Local\\Temp\\pear.exe-6b284e7e8d871a66f957486121bb627257c100641229989edb7f138792438755\\node_modules\\bare-crypto\\prebuilds\\win32-x64\\bare-crypto.bare'
}`

scripts/build-local-runtime.js
```diff
@@ -0,0 +1,21 @@
...
+const { spawnSync } = require('child_process')
+
+const root = path.resolve(__dirname, '..')
+const host = `${os.platform()}-${os.arch()}`
+const script = `make:${host}`
+const supported = new Set(['darwin-arm64', 'darwin-x64', 'linux-arm64', 'linux-x64', 'win32-x64'])
+
+if (!supported.has(host)) {
+  console.error(`Unsupported platform/arch: ${host}`)
+  console.error('Supported targets: darwin-arm64, darwin-x64, linux-arm64, linux-x64, win32-x64')
```
AndreiRegiani: Is it easy to add support for `win32-arm64` natively? it's missing there

jkcdarunday: @AndreiRegiani Should be easy since bare-build supports it
https://www.npmjs.com/package/bare-build-win32-arm64

subsystems/sidecar/lib/model.js
```diff
@@ -0,0 +1,135 @@
...
+      try {
+        await super.exit()
+      } finally {
+        this.#manual = false
+      }
+    }
+  }
+}
+
+class Model {
```
AndreiRegiani: Is this `model.js` really needed? this is the legacy pear-hyperdb collections being ported again here, instead can we remove it altogether? it covers legacy collections `traits`, `gc`, `presets` these were needed for `pear run` which is being removed already in this PR.

---

Just for knowledge, the new model class is here https://github.com/holepunchto/pear/blob/main/subsystems/sidecar/lib/db.js (uses `scripts/build-hyperdb.js"` -> generates `spec/`)

The new db has these collections: `dht`, `multisig`, all others aren't needed after pear run is removed.

scripts/build-local-runtime.js
AndreiRegiani: as a side note, the current prod pear only works on Windows ARM with Node-x64 installed (emulates transparently), if you happen to install Node-ARM (which is the correct one to choose) then it doesn't bootstrap since it doesn't exist in the prod drive, so it would be a nice small addition to actually support Windows ARM in the new pear-runtime architecture :)

geordangesink: yeah ok will add arm, I left it like this because wasn't at the point to support windows yet... am on it as we speak. (we need so adding to pear-runtime-updater)

subsystems/sidecar/lib/model.js
geordangesink: okay awesome, we can remove it then

rafapaezbas (approved):

PR #1091, opened 2026-05-26
Pear v3 refactor

rafapaezbas:


lib/state.js
jkcdarunday, line 1: We can remove this file I think

scripts/test.mjs
```diff
@@ -14,7 +14,7 @@ const root = path.dirname(dirname)
 const { default: checkout } = await import('../checkout')
```
jkcdarunday: We can remove this I think along with the global.Pear declaration below and the checkout.js file

jkcdarunday (approved):
lgtm

PR #1097, opened 2026-05-28
Support win32-arm64

AndreiRegiani:
`npm run make`
`npm run make:win32-arm64`

---

* Sub-task: solved this warning:

```bash
$ npm run make

(node:5052) [DEP0190] DeprecationWarning: Passing args to a child process with shell option true can lead to security vulnerabilities, as the arguments are not escaped, only concatenated.
(Use `node --trace-deprecation ...` to show where the warning was created)
```

---

image

### DIFF
--- a/scripts/link-win32-dev.js
+++ b/scripts/link-win32-dev.js
@@ -3,9 +3,10 @@
 
 const fs = require('fs')
 const path = require('path')
+const os = require('os')
 
 const root = path.resolve(__dirname, '..')
-const arch = process.argv[2]
+const arch = os.arch()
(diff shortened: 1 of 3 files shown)


scripts/link-win32-dev.js
```diff
@@ -5,7 +5,8 @@ const fs = require('fs')
 const path = require('path')
 
 const root = path.resolve(__dirname, '..')
-const runtimeRel = path.join('by-arch', 'win32-x64', 'bin', 'pear.exe')
+const arch = process.argv[2]
```
rafapaezbas: os.arch() should work hre

PR #1103, opened 2026-06-02
make channels

rafapaezbas:


constants.js
```diff
@@ -8,56 +8,83 @@ const b4a = require('b4a')
...
+  BOOT: require.main?.filename,
+  CONNECT_TIMEOUT: 20_000,
+  IDLE_TIMEOUT: 30_000,
+  SPINDOWN_TIMEOUT: 60_000,
+  KNOWN_NODES_LIMIT: 100,
+  get LOCALDEV() {
+    return !standalone
+  },
+  get UPGRADE() {
+    if (channel === null) throw new Error('UPGRADE read before setChannel()')
```
jkcdarunday: Hmm.. setChannel no longer seems to exist, should this be init instead?

scripts/make.js
```diff
@@ -0,0 +1,26 @@
...
+    '--host',
+    host,
+    '--out',
+    `./by-arch/${host}/bin`,
+    `targets/main.${channel}.js`
+  ],
+  { stdio: 'inherit', shell: true }
+)
+
+child.on('close', (code) => Bare.exit(code))
```
jkcdarunday: Best to handle signals like we do in test.mjs so we don't get passing CI despite the build actually failing. Also I think the event should be exit instead of close.
https://github.com/holepunchto/pear/blob/36549902a752120518e507523471c2fbc2488c0d/scripts/test.mjs#L58-L62

```suggestion
child.on('exit', (code, signal) => {
  Bare.exitCode = signal ? 128 + signal : code
})
```

jkcdarunday (changes requested):

constants.js
rafapaezbas: corrected

constants.js
```diff
@@ -8,56 +8,83 @@ const b4a = require('b4a')
...
+  CONNECT_TIMEOUT: 20_000,
+  IDLE_TIMEOUT: 30_000,
+  SPINDOWN_TIMEOUT: 60_000,
+  KNOWN_NODES_LIMIT: 100,
+  get LOCALDEV() {
+    return !standalone
+  },
+  get UPGRADE() {
+    if (channel === null) throw new Error('UPGRADE read before init()')
+    return pkg.upgrade[channel]
```
jkcdarunday: I'm thinking we could simplify the getter/setter setup for the entire constants.js.

Feels unnecessary to derive the platformUrl every time a value is read when the only point it changes is when init is called. Same goes for the UPGRADE getter here where it derives `pkg.upgrade[channel]` for every call.

I think it's fine to default the values to localdev and then update values directly in main.js since our only expected entrypoints are boot.js (localdev) and main.js (non-localdev)

What do you think?

jkcdarunday (approved):

constants.js
rafapaezbas: refactored to compute everything on init

PR #1106, opened 2026-06-03
v3 clean up --no-ask

AndreiRegiani:


.github/workflows/ci.yml
jkcdarunday: We can remove this now that we've added it as a separate CI job

jkcdarunday (changes requested):

.github/workflows/ci.yml
AndreiRegiani: Done. very true, now your npm test can pass green while lint fails separately

PR #1112, opened 2026-06-05
Add Signing

jkcdarunday:


package.json
```diff
@@ -14,7 +14,7 @@
     "test": "bare scripts/test.mjs",
     "format": "prettier --write . && lunte --fix",
     "lint": "prettier --check . && lunte",
-    "make": "bare scripts/make.js production",
```
rafapaezbas, line 17: I would keep production as default channel, seems like the safest option

jkcdarunday: Resolved in slack
https://holepunchdev.slack.com/archives/C054FE134P6/p1782477379928599?thread_ts=1782426625.820269&cid=C054FE134P6

It still defaults to production in script.js

jkcdarunday:
Signing and notarization successfully tested in:
https://github.com/holepunchto/pear-ci-build/actions/runs/28776422333/job/85321380641

rafapaezbas (approved):

PR #1131, opened 2026-07-08
Multithreaded vanity

jkcdarunday:


test/all.mjs
```diff
@@ -17,6 +17,7 @@ async function runTests() {
   await test.load(import.meta.resolve('./07-info.test.js'))
   await test.load(import.meta.resolve('./08-data.test.js'))
   await test.load(import.meta.resolve('./09-shutdown.test.js'))
+  await test.load(import.meta.resolve('./10-vanity.test.js'))
```
AndreiRegiani: shutdown test should be the last one always

AndreiRegiani: can add a //comment on it for future reminder

jkcdarunday: Thanks, moved it to the last. 

Though I'm curious, is there a technical reason for it or is it mainly for convention?

Right now shutdown tests run their own sidecars so its position shouldn't matter.

AndreiRegiani: @davidmarkclements do you remember? I guess it was coupled maybe became just a convention indeed

rafapaezbas: order of tests should not matter

rafapaezbas (approved):

PR #1136, opened 2026-07-14
Persist share debug logs

geordangesink:
Added --log-file <path> to persist runtime logs.
Persisted CLI and sidecar crash logs by default.
Added pear logs to inspect diagnostic log locations and files.
Added pear logs bundle [dir] to create a shareable log bundle.
Added pear logs clear to remove persisted logs.
Forwarded logging/debug options when spawning the sidecar.
Added JSON output support for log commands.


rafapaezbas:
Platform just needs to persist the sidecar logs in a known location

PR #1139, opened 2026-07-15
synchronize sidecar spindown tests

thundron:


test/10-shutdown.test.js
```diff
@@ -269,6 +270,32 @@ test('sidecar should not spindown until ongoing update is finished', async (t) =
...
+    expectSidecar: true
+  })
+  let sidecarClientClosed = false
+  t.teardown(async () => {
+    if (!sidecarClientClosed) await sidecarClient.close()
+  })
+  await sidecarClient.ready()
+  // Complete an operation so closing the client starts a fresh spindown countdown.
+  const touching = sidecarClient.touch({})
+  await Helper.pick(touching, { tag: 'final' })
```
jkcdarunday: I don't think that the touch part is necessary. Connecting and disconnecting a client should be enough.

https://github.com/holepunchto/pear/blob/8a8cf2f5a22d936397c20982f815ae5cf8a35749/subsystems/sidecar/index.js#L71-L82

test/10-shutdown.test.js
jkcdarunday: We can probably tighten this to like 3 seconds. The 10 seconds here was grace time to account for sidecar bootup and peer connection.

```suggestion
t.comment(
`Waiting for sidecar spindown timeout to lapse (${(SPINDOWN_TIMEOUT + 3_000) / 1000}s)`
)
const timeoutUntil = new Promise((resolve) => {
const timeout = setTimeout(() => resolve(false), SPINDOWN_TIMEOUT + 3_000)
```

jkcdarunday (changes requested):

thundron:
@jkcdarunday looks like `touch` is needed for windows, as it internally does `await sidecar.ready()`

jkcdarunday:
Hmm.. That's odd because `sidecar.ready()` is also called in the constructor. And sidecar should still connect to swarm and update even without any operation.

https://github.com/holepunchto/pear/blob/8a8cf2f5a22d936397c20982f815ae5cf8a35749/subsystems/sidecar/index.js#L91-L96

thundron:
> Hmm.. That's odd because `sidecar.ready()` is also called in the constructor. And sidecar should still connect to swarm and update even without any operation.

hmmm checked again, it looks like it's just the global timeout, much like https://github.com/holepunchto/pear/pull/1118/changes#diff-1e2f2f263fc1cf5094419d8710e81167ae3baedc981136a552b66b159c25fd6aR9 in the other PR I have

test/10-shutdown.test.js
```diff
@@ -269,19 +270,48 @@ test('sidecar should not spindown until ongoing update is finished', async (t) =
...
-  )
+  // Keep the updater sidecar alive until peer discovery completes.
+  const sidecarClient = new Helper({
+    platformDir: path.join(buildDir, 'pear'),
+    expectSidecar: true
+  })
+  let sidecarClientClosed = false
+  t.teardown(async () => {
+    if (!sidecarClientClosed) await sidecarClient.close()
+  })
```
jkcdarunday: We don't need the sidecarClientClosed here I think. ReadyResource already has a check to prevent double-close https://github.com/holepunchto/ready-resource/blob/5009db6e91fe596067d1b355f7ca5a04be254d2c/index.js#L21 which is also cleaner because it awaits the existing close instead of completing the teardown immediately
```suggestion
  t.teardown(() => sidecarClient.close())
```

test/10-shutdown.test.js
```diff
@@ -269,19 +270,48 @@ test('sidecar should not spindown until ongoing update is finished', async (t) =
...
+  const peerAddedTimeout = new Promise((resolve) => {
+    peerAddedTimer = setTimeout(() => resolve(false), 30_000)
+  })
+  const peerAdded = await Promise.race([
+    peerAddedUntil,
+    untilExit.then(() => false),
+    peerAddedTimeout
+  ])
+  clearTimeout(peerAddedTimer)
+  t.is(peerAdded, true, 'sidecar successfully connected to throttled seeder')
```
jkcdarunday:
```suggestion
  t.ok(peerAdded, 'sidecar successfully connected to throttled seeder')
```

test/10-shutdown.test.js
```diff
@@ -269,19 +270,48 @@ test('sidecar should not spindown until ongoing update is finished', async (t) =
...
+
+  await sidecarClient.close()
+  sidecarClientClosed = true
+
+  if (!peerAdded) {
+    t.fail('cannot test spindown before updater connects')
+    return
+  }
```
jkcdarunday: The t.fail here kinda duplicates the check at 295. 
```suggestion
  if (!peerAdded) return
```

test/10-shutdown.test.js
```diff
@@ -269,19 +270,48 @@ test('sidecar should not spindown until ongoing update is finished', async (t) =
   const untilExit = new Promise((resolve) => sidecar.once('exit', resolve))
   t.teardown(() => untilExit, { order: Infinity })
 
-  t.comment(
-    `Waiting for sidecar spindown timeout to lapse (${(SPINDOWN_TIMEOUT + 10_000) / 1000}s)`
-  )
+  // Keep the updater sidecar alive until peer discovery completes.
+  const sidecarClient = new Helper({
```
jkcdarunday: Instead of a comment, maybe we can name this keepAliveClient
```suggestion
  const keepAliveClient = new Helper({
```

jkcdarunday (changes requested):

PR #1149, opened 2026-07-22
make link mandatory in pear gc cores

thundron:


test/13-gc.test.js
```diff
@@ -0,0 +1,87 @@
...
+  t.teardown(() => helper.close(), { order: Infinity })
+  await helper.ready()
+
+  const target = await createDrive(t)
+  const other = await createDrive(t)
+  await cacheDrive(t, helper, target.link)
+  await cacheDrive(t, helper, other.link)
+
+  const removed = []
+  const collecting = helper.gc({ resource: 'cores', data: { link: target.link } })
```
RaisinTen, line 21: Should we also test here that `helper.cores()` isn't returning the gc-ed pear links?

thundron: yes, sounds more than reasonable

RaisinTen (dismissed):

rafapaezbas (approved):

PR #1161, opened 2026-07-29
Handle pear-build and OPERATION_FAILED errors and unify pear link parsing

jkcdarunday:
Changes:
- Add handling of non-critical pear errors
- Extract actual error from ERR_OPERATION_FAILED errors
- Unify validation/parsing of pear links


AndreiRegiani:
Is the stack trace expected to show here? cc @jkcdarunday 

image

lib/crasher.js
```diff
@@ -1,15 +1,30 @@
...
+  PearError.ERR_OPERATION_FAILED
+])
+
 function setupCrashHandlers(label) {
   const runContext = global.Bare
 
   runContext.on('unhandledRejection', (err) => {
-    LOG.error(label, 'Exiting due to unhandled rejection', err)
+    if (!(err instanceof PearError) || unhandled.has(err.constructor)) {
+      LOG.error(label, 'Exiting due to unhandled rejection', err)
```
rafapaezbas: I dont understan this change, why dont we log?

cmd/changelog.js
```diff
@@ -20,9 +20,7 @@ module.exports = async function changelog(cmd) {
   const ipc = context.getIPC()
   const { json, full, max = 10 } = cmd.flags
   const link = cmd.args.link || null
-  if (link && plink.parse(link).drive.key === null) {
-    throw ERR_INVALID_INPUT('Link "' + link + '" is not a valid key')
-  }
+  parse(link)
```
AndreiRegiani: `link=null` this breaks `$ pear changelog #nolink` -> Shows Pear/Platform changelog by default

```bash
./pear.dev changelog
✖ A valid pear link must be specified.
```

jkcdarunday: Nice catch, thanks! Fixed for both changelog and gc.

lib/crasher.js
jkcdarunday: After rechecking, it's actually hitting this path because pear-build both rejects and emits the error. The emit causes an unhandled rejection instead which leads to this path.

I don't think we need this change anymore as we should ignore the emit instead. Will remove.

AndreiRegiani:
looks like this PR caused behavior changes, loss of features regarding `file://` args, will it be dropped?

main branch

image

this PR

image

AndreiRegiani:
another behavior change / bug, `pear build` doesn't show help page anymore:

---

main branch

image

this PR

image

maybe it's missing something like this `if (!cmd.flags.package) return console.log(build.help())`

rafapaezbas (approved):

jkcdarunday:
> another behavior change / bug, `pear build` doesn't show help page anymore:

@AndreiRegiani 
Ah it's still missing the pear-build PR at https://github.com/holepunchto/pear-build/pull/26

PR #1201, opened 2026-08-25
Implement pear blind-relay start

RaisinTen:
https://github.com/user-attachments/assets/52792e26-dbae-436b-8515-1190041e660b

https://github.com/user-attachments/assets/0d4c1e5a-c6bd-44b9-a0a5-699fe105d3d2

<details>

<summary>Client</summary>

```js
'use strict'

const DHT = require('hyperdht')
const BlindRelay = require('blind-relay')
const hypercoreid = require('hypercore-id-encoding')

async function main () {
  const publicKeyZ32 = process.argv[2]

  if (!publicKeyZ32) {
    console.error('usage: node client.js <relay-public-key-z32>')
    process.exit(1)
  }

  const relayPublicKey = hypercoreid.decode(publicKeyZ32)

  const dht = new DHT()

  console.log('connecting to relay:', publicKeyZ32)

  const socket = dht.connect(relayPublicKey)

  const client = new BlindRelay.Client(socket)

  client.on('open', () => {
    console.log('connected')
  })

  client.on('close', () => {
    console.log('disconnected')
  })

  client.on('error', (err) => {
    console.error('client error:', err)
  })

  // Keep the process alive.
  await new Promise(() => {})
}

main().catch(console.error)
```

</details>

### DIFF
--- a/cmd/blind-relay.js
+++ b/cmd/blind-relay.js
@@ -111,7 +111,8 @@ module.exports = async function blindRelay(cmd) {
   const layout = new TableLayout(
     [
       { type: 'border', char: ' ' },
-      { type: 'table', table: stats }
+      { type: 'table', table: stats },
+      { type: 'border', char: '─' }
     ],
     { appendMode: !ctrlTTY }
   )
--- a/cmd/blind-relay.js
+++ b/cmd/blind-relay.js
@@ -1,5 +1,6 @@
 'use strict'
 const context = require('../context')
+const hypercoreid = require('hypercore-id-encoding')
 
 const { ERR_INVALID_INPUT } = require('pear-errors')
 const { outputter, ansi, isTTY, stdio, setupTTYInput } = require('../lib/terminal.js')
@@ -139,7 +140,7 @@ module.exports = async function blindRelay(cmd) {
       }
     },
     listening: ({ publicKey }) => {
-      stats.update({ publicKey })
+      stats.update({ publicKey: hypercoreid.normalize(publicKey) })
       layout.print(stdio)
       // return `Blind-relay listening on ${publicKey}`
     },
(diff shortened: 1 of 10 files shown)


cmd/blind-relay.js
```diff
@@ -0,0 +1,219 @@
...
+  ])
+  const layout = new TableLayout(
+    [
+      { type: 'border', char: ' ' },
+      { type: 'table', table: stats }
+    ],
+    { appendMode: !ctrlTTY }
+  )
+
+  if (ctrlTTY) {
```
rafapaezbas: I think this logic exists also in cmd/seed, can we re-use the code?

subsystems/sidecar/ops/blind-relay.js
```diff
@@ -0,0 +1,82 @@
...
+      LOG.trace('blind-relay', 'connection accepted', {
+        remotePublicKey: z32.encode(socket.remotePublicKey)
+      })
+
+      return this.relay.accept(socket, {
+        id: socket.remotePublicKey
+      })
+    })
+
+    await this.server.listen(this.keyPair)
```
rafapaezbas: what happens if user starts two servers? maybe we should keep state in the sidecar.js?

cmd/blind-relay.js
AndreiRegiani: >re-use the code

agreed

cmd/index.js
```diff
@@ -521,6 +521,26 @@ module.exports = async (ipc, argv = cmdArgs) => {
     commands.cores
   )
 
+  const blindRelay = command(
+    'blind-relay',
+    summary('Run a blind relay that forwards UDX streams between peers that cannot hole-punch'),
```
AndreiRegiani: description is too long

image

rafapaezbas (approved):

PR #1212, opened 2026-09-04
fix blind relay

rafapaezbas:


subsystems/sidecar/ops/blind-relay.js
```diff
@@ -15,10 +15,11 @@ class BlindRelayServer extends ReadyResource {
   async _open() {
     LOG.trace('blind-relay', 'starting')
 
+    const dht = this.dht
     this.relay = new RelayServer({
       createStream(opts) {
         LOG.trace('blind-relay', 'creating raw stream')
-        return this.dht.createRawStream({ ...opts, framed: true })
+        return dht.createRawStream({ ...opts, framed: true })
```
RaisinTen: Should we use an arrow-function here? It seems more idiomatic to capture `this`:
```suggestion
this.relay = new RelayServer({
createStream: (opts) => {
LOG.trace('blind-relay', 'creating raw stream')
        return this.dht.createRawStream({ ...opts, framed: true })
```

RaisinTen (dismissed):

RaisinTen (approved):

